PR #8, opened 2015-11-11
Added IRC nick colors

Temeez:
This greatly increases the readability in IRC channels (at least when using irssi).


Temeez:
I'll fix the checks and make a new one.

PR #12, opened 2015-11-13
Add Direct Passthrough Users

dkabot:
Being a rebel, I plan to see if I can use Discord-IRC to basically set up Discord as a personal, nice IRC client. In such a case, the <user> before every message becomes redundant and annoying. Since just ripping out that logic in a personal fork seemed like a bad idea (what if someone else *does* happen to connect and type something? I don't want to be impersonated!) I made this instead.

Discord user IDs can be added to an optional "passthroughUserIDs" section of the configuration, and if that user sends a message, it just sends through the bot directly. No <username>, no "Command sent by X from Discord", they just talk through the bot. If it doesn't exist in the config, or the user ID isn't in it, it acts as usual.

While what I'm trying to do isn't the goal of the project, I thought this might happen to have its uses. This is definitely a very niche thing to have, but since it works I figured why not PR it. My apologies if this is too out there.

(I'm definitely not a node.js dev, so I might have missed an easier way to do this, though the `npm run lint` didn't output anything. Additional apologies if there's anything obviously off.)

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -144,9 +144,9 @@ class Bot {
       const colorIndex = (username.charCodeAt(0) + username.length) % NICK_COLORS.length;
       const coloredUsername = irc.colors.wrap(NICK_COLORS[colorIndex], username);
       let text = this.parseText(message);
-      
+
       // Don't add a username when it's a passthrough user
-      if(!this.isPassthroughUser(author)) {
+      if (!this.isPassthroughUser(author)) {
         if (this.isCommandMessage(text)) {
           const prelude = `Command sent from Discord by ${username}:`;
           this.ircClient.say(ircChannel, prelude);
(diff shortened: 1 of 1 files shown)


ekmartin:
Hi, and thanks for the pull request! This definitely looks like a clean solution, but as you're saying I'm a bit unsure if it's within the project's goal. I feel like if you want to use Discord as an IRC client, there's much cooler ways to do it. 

For example, what happens to private messages? Right now they'll just get absorbed, and you'll never know they arrived. If you had a private Discord server as your IRC bridge you could instead print them out, for example with a prefix that shows that it's a private message, and who they're from. Then you could also make it possible to send PMs to people on IRC, maybe through a specific command sent to the Discord bot user (i.e. `!message <username> hi`).

You could also implement other special commands, like for listing all the users in the IRC channel, showing the message of the day and so on.

If this is something you'd want to spend time building I wouldn't mind at all if you used discord-irc as a basis, i.e. by creating a new repository with a different name, but with discord-irc's existing code.

ekmartin:
Yep, I see. Feel free to PM me with questions on Discord (ekmartin) if you give it a try in the future :smile:

PR #18, opened 2015-12-22
Enabling multi-server support for Discord

billimek:
configure channel mapping such as,
"channelMapping": {
   "DiscordServerName,#other-discord": "#new-irc-channel"
}

I haven't fixed all of the tests, still learning, but want to give you a pull request now for the actual implementation patch

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Create a discord user and join the server you want to connect to IRC.
     "discordEmail": "user@mail.com",
     "discordPassword": "letmein123",
     "channelMapping": {
-      "#other-discord": "#new-irc-channel"
+      "DiscordServerName,#other-discord": "#new-irc-channel"
     }
   },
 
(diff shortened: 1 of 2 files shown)


billimek:
Ok so The code changes to the actual application took about 30 mins to implement and test (live testing).  I've spent about 4-5 hours trying to get the test harness to work with the changes and am at a big loss.  I updated all of the fixtures to represent the new channelMapping syntax in the configs as well as changing the appropriate expected results.  

However I ran into roadblocks trying to figure out what was required to change in (primarily) bot.test.js to make this work.

I believe there is a problem with the new change in lib/bot.js, `const discordChannel = this.discord.servers.get('name', discordServerName).channels.get('name', discordChannelName.slice(1));` where servers.get() is returning an object that is the subsequent parent to channels.  I couldn't figure out how to make the test stuff work with this.  Tried many many different things with the test framework and stubs to no success.

ekmartin:
Everything about Discord is mocked away in the tests, so changes like these can be a bit painful. The stub for Discord does not have a `servers` field, which it would need: https://github.com/reactiflux/discord-irc/blob/master/test%2Fstubs%2Fdiscord-stub.js#L19

Could be something like:
```js
this.servers = {
  get() {
    return this; // since DiscordStub has a `channels` method. 
    // another option would be to create a new ServerStub and return that
  }
}
```

If you don't have time I can update the tests for you before merging in the changes. I'm a bit unsure about the comma separated configuration syntax though. Maybe the channel mapping values could be given either as an object or a string instead, example:

```json
"channelMapping": {
   "#discord": {
     "server": "Server",
     "channel": "#irc"
   },
   "#other-discord": "#other-irc"
}
```

That would make the change backwards compatible too.

PR #19, opened 2015-12-23
IRC Nickname color toggle

Bryceclark56:
A quick workaround for people that might not like the colors in IRC.
Adds a config option to enable IRC colors

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,8 +64,8 @@ Create a discord user and join the server you want to connect to IRC.
     "ircOptions": { // Optional node-irc options
       "floodProtection": false, // On by default
       "floodProtectionDelay": 1000, // 500 by default
-      "nickColor": true // Off by default
     },
+    ircNickColor: false, // Gives usernames a color in IRC for better readability (off by default)
     // Makes the bot hide the username prefix for messages that start
     // with one of these characters (commands):
     "commandCharacters": ["!", "."]
--- a/README.md
+++ b/README.md
@@ -63,7 +63,8 @@ Create a discord user and join the server you want to connect to IRC.
     },
     "ircOptions": { // Optional node-irc options
       "floodProtection": false, // On by default
-      "floodProtectionDelay": 1000 // 500 by default
+      "floodProtectionDelay": 1000, // 500 by default
+      "nickColor": true // Off by default
     },
     // Makes the bot hide the username prefix for messages that start
     // with one of these characters (commands):
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -138,13 +138,17 @@ class Bot {
       const colorIndex = (username.charCodeAt(0) + username.length) % NICK_COLORS.length;
       const coloredUsername = irc.colors.wrap(NICK_COLORS[colorIndex], username);
       let text = this.parseText(message);
+      let newUsername = username;
+      if (this.ircOptions.nickColor) {
+        newUsername = coloredUsername;
+      }
 
       if (this.isCommandMessage(text)) {
         const prelude = `Command sent from Discord by ${username}:`;
         this.ircClient.say(ircChannel, prelude);
         this.ircClient.say(ircChannel, text);
       } else {
-        text = `<${coloredUsername}> ${text}`;
+        text = `<${newUsername}> ${text}`;
         logger.debug('Sending message to IRC', ircChannel, text);
         this.ircClient.say(ircChannel, text);
       }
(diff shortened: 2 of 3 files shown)


lib/bot.js
```diff
@@ -138,13 +138,17 @@ class Bot {
       const colorIndex = (username.charCodeAt(0) + username.length) % NICK_COLORS.length;
```
ekmartin: Maybe put these two lines into the `if` too? No need to create `colorIndex` and `coloredUsername` if they aren't gonna be used.

ekmartin:
Cool! However, `ircOptions` are passed directly to node-irc (could've had a better name). Would be better to put this as a root option, and name it something like `ircNickColor`.

README.md
```diff
@@ -63,8 +63,9 @@ Create a discord user and join the server you want to connect to IRC.
     },
     "ircOptions": { // Optional node-irc options
       "floodProtection": false, // On by default
-      "floodProtectionDelay": 1000 // 500 by default
+      "floodProtectionDelay": 1000, // 500 by default
     },
+    ircNickColor: false, // Gives usernames a color in IRC for better readability (off by default)
```
ekmartin: You forgot quotes around `ircNickColor` :smiley:. (`ircNickColor` -> `"ircNickColor"`)

lib/bot.js
```diff
@@ -31,6 +31,7 @@ class Bot {
     this.discordEmail = options.discordEmail;
     this.discordPassword = options.discordPassword;
     this.commandCharacters = options.commandCharacters || [];
+    this.ircNickColor = options.ircNickColor || false;
```
ekmartin: Should this maybe default to true? Since that's the existing behavior.

lib/bot.js
```diff
@@ -135,16 +136,20 @@ class Bot {
     logger.debug('Channel Mapping', channelName, this.channelMapping[channelName]);
     if (ircChannel) {
       const username = author.username;
-      const colorIndex = (username.charCodeAt(0) + username.length) % NICK_COLORS.length;
-      const coloredUsername = irc.colors.wrap(NICK_COLORS[colorIndex], username);
       let text = this.parseText(message);
+      let newUsername = username;
+      if (this.ircNickColor) {
+        const colorIndex = (username.charCodeAt(0) + username.length) % NICK_COLORS.length;
+        const coloredUsername = irc.colors.wrap(NICK_COLORS[colorIndex], username);
```
ekmartin: Don't need coloredUsername here, can just do `newUsername = irc.colors.wrap...`. I think something like `displayUsername` would be a better variable name than `newUsername` though :)

ekmartin:
Would you want to add a test for disabling colors? [This](https://github.com/reactiflux/discord-irc/blob/master/test%2Fbot.test.js#L67-L85) is a good starting point. If you don't have time I can fix it when merging.

README.md
```diff
@@ -63,8 +63,9 @@ Create a discord user and join the server you want to connect to IRC.
     },
     "ircOptions": { // Optional node-irc options
       "floodProtection": false, // On by default
-      "floodProtectionDelay": 1000 // 500 by default
+      "floodProtectionDelay": 1000, // 500 by default
```
ekmartin: Why the trailing comma here?

README.md
```diff
@@ -63,8 +63,9 @@ Create a discord user and join the server you want to connect to IRC.
...
     },
+    "ircNickColor": false, // Gives usernames a color in IRC for better readability (on by default)
```
ekmartin: Maybe this should be an object, so that you could toggle colors for both IRC and Discord?
Something like this:
```js
"nickColors": {
  "irc": true,
  "discord": false
}
```

Or maybe only one option that covers both is necessary? I.e. something like `nickColors`.

lib/bot.js
```diff
@@ -135,16 +136,18 @@ class Bot {
     logger.debug('Channel Mapping', channelName, this.channelMapping[channelName]);
     if (ircChannel) {
       const username = author.username;
-      const colorIndex = (username.charCodeAt(0) + username.length) % NICK_COLORS.length;
-      const coloredUsername = irc.colors.wrap(NICK_COLORS[colorIndex], username);
       let text = this.parseText(message);
+      let displayUsername = username;
+      if (this.ircNickColor) {
+        displayUsername = irc.colors.wrap(NICK_COLORS[(username.charCodeAt(0) + username.length) % NICK_COLORS.length], username);
```
ekmartin: Since we have lodash available you could use `_.sample(NICK_COLORS)`: https://lodash.com/docs#sample

ekmartin:
Woops! Misread a bit, will add a test and merge this. Thanks!

ekmartin:
Released in version 0.7.0, thanks! Set the default to true instead of false, to not break backwards compatibility.

PR #32, opened 2016-02-04
Send attachment URL to IRC

### LINKED ISSUE #14: Sends empty message to IRC when an attachment is added
Not sure how robust discord.js is but when someone attaches a file of any format it will just send a blank message to IRC. It would be cool to pull the name of the attachment and put something like "<USER> Attached file 'filename.ext'." if that's not possible maybe just have it say something like "<USER> Attached a file, please log into Discord to view it."

rce:
Also closes #14.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -145,7 +145,8 @@ class Bot {
         this.ircClient.say(ircChannel, text);
       } else if (message.attachments && message.attachments.length) {
         message.attachments.forEach(a => {
-          const urlMessage = `<${coloredUsername}> ${a.url}`;
+          const urlMessage =
+            `${coloredUsername} posted an attachment to ${channelName} on Discord: ${a.url}`;
           logger.debug('Sending message to IRC', ircChannel, urlMessage);
           this.ircClient.say(ircChannel, urlMessage);
         });
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -144,9 +144,11 @@ class Bot {
         this.ircClient.say(ircChannel, prelude);
         this.ircClient.say(ircChannel, text);
       } else {
-        text = `<${coloredUsername}> ${text}`;
-        logger.debug('Sending message to IRC', ircChannel, text);
-        this.ircClient.say(ircChannel, text);
+        if (text !== '') {
+          text = `<${coloredUsername}> ${text}`;
+          logger.debug('Sending message to IRC', ircChannel, text);
+          this.ircClient.say(ircChannel, text);
+        }
         if (message.attachments) {
           message.attachments.forEach(a =>
             this.ircClient.say(ircChannel, `<${coloredUsername}> ${a.url}`));
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -143,19 +143,16 @@ class Bot {
         const prelude = `Command sent from Discord by ${username}:`;
         this.ircClient.say(ircChannel, prelude);
         this.ircClient.say(ircChannel, text);
+      } else if (message.attachments && message.attachments.length) {
+        message.attachments.forEach(a => {
+          const urlMessage = `<${coloredUsername}> ${a.url}`;
+          logger.debug('Sending message to IRC', ircChannel, urlMessage);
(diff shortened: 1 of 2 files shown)


lib/bot.js
```diff
@@ -144,9 +144,18 @@ class Bot {
...
       } else {
-        text = `<${coloredUsername}> ${text}`;
-        logger.debug('Sending message to IRC', ircChannel, text);
-        this.ircClient.say(ircChannel, text);
+        if (text !== '') {
+          text = `<${coloredUsername}> ${text}`;
+          logger.debug('Sending message to IRC', ircChannel, text);
+          this.ircClient.say(ircChannel, text);
+        }
+        if (message.attachments) {
```
ekmartin: Can a message have both an attachment and a message? If not, maybe it would be better to do 

```js
if (message.attachments) { 
  // something
} else {
  // send regular message
}

File: lib/bot.js [hunk 2]
```diff
@@ -144,9 +144,18 @@ class Bot {
...
+          message.attachments.forEach(a => {
+            const urlMessage = `<${coloredUsername}> ${a.url}`;
```
ekmartin: Maybe it would be better to post a more human friendly description here?
Something like: `Bob has posted an attachment to #channel on Discord: <link>` or similar.

lib/bot.js
```diff
@@ -144,9 +144,18 @@ class Bot {
...
+            logger.debug('Sending message to IRC', ircChannel, urlMessage);
```
ekmartin: Could update this so it logs something else, makes it easier to debug :)

ekmartin:
Nice, thanks! Would be really good with some tests though, if you've got the time.

lib/bot.js
rce: So far I haven't found a way to get both into a single message. I'll do as you suggested!

lib/bot.js
rce: Good point, will do.

rce:
I'll see what I can do about the tests. Only realised afterwards you already had some message tests in place so it should be pretty quick to do.

lib/bot.js
```diff
@@ -143,7 +143,14 @@ class Bot {
...
         this.ircClient.say(ircChannel, text);
-      } else {
+      } else if (message.attachments) {
+        message.attachments.forEach(a => {
+          const urlMessage =
+            `${coloredUsername} posted an attachment to ${channelName} on Discord: ${a.url}`;
+          logger.debug('Sending attachment URL to IRC', ircChannel, urlMessage);
+          this.ircClient.say(ircChannel, urlMessage);
+        });
+      } else if (text !== '') {
```
ekmartin: This should probably just be `} else {`.

rce:
Should be pretty good now (with tests!) unless I crated more issues, hah!

lib/bot.js
```diff
@@ -143,6 +143,13 @@ class Bot {
         const prelude = `Command sent from Discord by ${username}:`;
         this.ircClient.say(ircChannel, prelude);
         this.ircClient.say(ircChannel, text);
+      } else if (message.attachments) {
```
ekmartin: Maybe it would be smart to check if there actually are any attachments too? I.e.:
`} else if (message.attachments && message.attacments.length) {`.

Makes sure it doesn't bug if a message includes an empty attachments array.

ekmartin:
Other than that it looks great!

lib/bot.js
rce: It shouldn't cause issues beacuse `message.attachments.forEach(...)` is used.

ekmartin: You won't reach the else statement though, so nothing will actually happen with the message.

I guess it's fine if Discord doesn't include an empty attachments array for normal messages, but might be good to have nonetheless. What do you think?

rce: That's a great point. I'll add the check.

ekmartin:
Thanks!

ekmartin:
Released in version 0.5.0.

PR #33, opened 2016-02-24
Fix sending attachments to IRC

rce:
Now that Discord actually got the ability to send both text messages and attachments in single message, here's the code to support that.

This time with tests right off the bat!

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -148,9 +148,11 @@ class Bot {
         this.ircClient.say(ircChannel, prelude);
         this.ircClient.say(ircChannel, text);
       } else {
-        text = `<${displayUsername}> ${text}`;
-        logger.debug('Sending message to IRC', ircChannel, text);
-        this.ircClient.say(ircChannel, text);
+        if (text !== '') {
+          text = `<${displayUsername}> ${text}`;
+          logger.debug('Sending message to IRC', ircChannel, text);
+          this.ircClient.say(ircChannel, text);
+        }
 
         if (message.attachments && message.attachments.length) {
           message.attachments.forEach(a => {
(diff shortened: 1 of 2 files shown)


rce:
I wonder what went wrong with the frigg build? Results look correct other than the return code `255` which I can't reproduce on my machine.

ekmartin:
Not sure, seems like it passed after a restart. Would it be better to make it clearer that the message actually is a comment to the attachment? Something like "ekmartin posted an attachment to #general on Discord, with the comment "here's the file": <url> - or something like that?

ekmartin:
Any update on this? :)

rce:
I've been running a few variations of this myself because I wasn't sure what would be the best format to post the attachments and comment to IRC.

We have ended up with sending the comment and attachment link(s) as if they were simple normal messages like this:

```
<rce> So good cola!
<rce> https://cdn.discordapp.com/attachments/141649840923869184/164022249014165506/IMG_0620.JPG
```

Instead of this

```
<rce> So good cola!
rce posted an attachment to #channel on Discord: https://cdn.discordapp.com/attachments/141649840923869184/164022249014165506/IMG_0620.JPG
```

or this

```
rce posted an attachment to #channel on Discord, with the comment "So good cola!": https://cdn.discordapp.com/attachments/141649840923869184/164022249014165506/IMG_0620.JPG
```

We found that the extra text "X posted an attachment to Y on Discord" is very much noise especially if the channel receives attachments often. The links itself are pretty damn long by themselves. None of the other messages to IRC include similar extra info either. What the IRC user wants to see is just the comment and link to attachment anyway.

FWIW Discord doesn't distinguish between an attachment with comment and a text message and attachment sent separately.

image

What do you think?

ekmartin:
Sounds good, your points about too much noise makes sense. Care to rebase with master after changing to what you're suggesting? So the branch is mergeable :)

rce:
Done!

ekmartin:
Great, thanks!

ekmartin:
Published as [version 0.8.0](https://github.com/reactiflux/discord-irc/blob/master/CHANGELOG.md#080---2016-04-04).

PR #57, opened 2016-04-21
Add options to configure Discord messages

ajwhite:
This allows `**<author>**` to be configurable when sending a message from IRC to Discord.

A `discordOptions` object is added to the config with an `includeAuthor` property. This is enabled by default, so there are no BC changes, it just allows for it to be excluded.

As for the implementation, I moved the formatting implementation details into a new `formatDiscordMessage` function since there is additional logic, and the potential for further rules to become applied.

----

I want to use `discord-irc` to allow myself to use Discord _as a user_ through IRC. Rather than having the IRC chat be a proxy for many users, through the Discord bot user, I want to treat it more as a direct proxy for _my_ Discord account. 

All I need is for messages from IRC to omit the author tag and I'm good to go!

### Example

- `atticus-test` is "another user in the Discord chat"
- `atticus` is me, both my Discord user, and the person chatting through IRC. 

screen shot 2016-04-21 at 12 02 21 am

screen shot 2016-04-21 at 12 00 55 am

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,9 @@ Create a discord user and join the server you want to connect to IRC.
     "channelMapping": { // Maps each Discord-channel to an IRC-channel, used to direct messages to the correct place
       "#discord": "#irc channel-password" // Add channel keys after the channel name
     },
+    "discordOptions": { // Optional Discord message options
+      "includeAuthor": false // On by default
+    },
     "ircOptions": { // Optional node-irc options
       "floodProtection": false, // On by default
       "floodProtectionDelay": 1000 // 500 by default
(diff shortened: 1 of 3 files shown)


TheDoctorsLife:
This is a cool feature.

ekmartin:
Sorry for the late answer. Really clean implementation and PR, but I'm a bit unsure about the feature itself. Something similar (albeit the opposite) has been discussed before, here: https://github.com/reactiflux/discord-irc/pull/12

Specifically, see my comment here: https://github.com/reactiflux/discord-irc/pull/12#issuecomment-156758604

I've discussed this a bit over at https://github.com/ekmartin/slack-irc too, which is what discord-irc is based on. In the end I decided to rather recommend https://github.com/aeirola/slack-irc-client - which lets you use Slack as a personal IRC client (with PM support and similar).

I'm not completely sure though, but I think this might be a bit of a halfway measure compared to what you could actually get working if you didn't care about preserving the rest of the core functionality in discord-irc. What do you think?

PR #81, opened 2016-06-26
Fix pings to users with nicknames showing as raw user IDs in IRC

GinjaNinja32:
before: `<GinjaNinja32> <@!169859930382270465>`
after: `<GinjaNinja32> @GinjaNinja32`

Both of these show as `@Ginja` on the Discord side; given discord-irc already ignores nicknames for the user that's shown on IRC, I decided against replacing the user's nickname in instead for `<@! ... >` pings.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -111,6 +111,7 @@ class Bot {
   parseText(message) {
     const text = message.mentions.reduce((content, mention) => (
       content.replace(`<@${mention.id}>`, `@${mention.username}`)
+             .replace(`<@!${mention.id}>`, `@${mention.username}`)
     ), message.content);
 
     return text
--- a/test/bot.test.js
+++ b/test/bot.test.js
@@ -238,6 +238,18 @@ describe('Bot', function() {
     this.bot.parseText(message).should.equal('@testuser hi');
   });
 
+  it('should convert user nickname mentions from discord', function() {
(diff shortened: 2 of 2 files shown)


lib/bot.js
```diff
@@ -111,6 +111,7 @@ class Bot {
   parseText(message) {
     const text = message.mentions.reduce((content, mention) => (
       content.replace(`<@${mention.id}>`, `@${mention.username}`)
+             .replace(`<@!${mention.id}>`, `@${mention.username}`)
```
ekmartin: Cool! Maybe the line before isn't needed anymore then (maybe Discord changed something)?

GinjaNinja32: Users *without* nicknames are pinged as `<@id>`, users with are pinged as `<@!id>`. Not sure why there needs to be a difference, but there is one. Both lines are needed.

ekmartin: Ah, I see! Would you mind adding a test as well?
Should be pretty identical to this one: https://github.com/reactiflux/discord-irc/blob/master/test/bot.test.js#L229-L239

(You run the tests with npm test)

GinjaNinja32: Test added.

ekmartin:
Great, thanks! I'll publish a new release later today.

Issue #89, opened 2016-07-28
IRC Color code


Dommage:
Any way to have discord-irc to remove color code from message?

ekmartin:
This has been fixed since through the config option `"ircNickColor": false`.

Issue #86, opened 2016-07-21
Imperfect Discord text handling


FlashingYoshi:
1. Discord to IRC messages will cut off and not extend if they run on for too long (sometimes).
2. Both Discord to IRC and IRC to Discord messages lack the proper formatting.
3. IRC do not send bot-prefixed messages to Discord as one would expect.

DarkSpyro003:
1. should be fixed by #129 
I'm not sure what is meant by 3.

rahatarmanahmed:
2 should be covered once #205 is merged. Believe 1 and 3 are already supported so we can probably close this.

PR #97, opened 2016-08-04
Added formatting conversions to messages

Kurocon:
I added text formatting conversions to messages. Things like bold, italic and underscored text will be converted into their counterparts on the other service, and colors and strikethroughs will be stripped because they are unsupported on the other platform.

This would fix number 2 from issue #86 and issue #89.

### For example on discord:
discord-irc_formatting1

### And on IRC:
discord-irc_formatting2

### DIFF
--- a/lib/formatting.js
+++ b/lib/formatting.js
@@ -55,7 +55,7 @@ export function formatFromIRCToDiscord(text) {
   }
 
   // Remove all color markers from the message, as discord does not support colored text
-  resultString = resultString.replace(/\u0003(\d\d(,\d\d)?)?/g, '');
+  resultString = resultString.replace(/\u0003(\d\d?(,\d\d?)?)?/g, '');
 
   return resultString;
 }
--- a/lib/formatting.js
+++ b/lib/formatting.js
@@ -1,0 +1,58 @@
+/* eslint-disable no-unused-vars */
+
+export function formatFromDiscordToIRC(text) {
+  // Apply formatting rules, from most specific to least specific.
+  return text.replace(/~~(.*?)~~/g, '$1')                        // Remove unsupported strikethrough
+    .replace(/__\*\*\*(.*?)\*\*\*__/g, '\u001F\u001D\u0002$1\u000F') // Apply underline-italics-bold
+    .replace(/__\*\*(.*?)\*\*__/g, '\u001F\u0002$1\u000F')           // Apply underline-bold
+    .replace(/__\*(.*?)\*__/g, '\u001F\u001D$1\u000F')               // Apply underline-italics
+    .replace(/__(.*?)__/g, '\u001F$1\u000F')                         // Apply underline
+    .replace(/\*\*\*(.*?)\*\*\*/g, '\u0002\u001D$1\u000F')          // Apply bold-italics
+    .replace(/\*\*(.*?)\*\*/g, '\u0002$1\u000F')                     // Apply bold
+    .replace(/\*(.*?)\*/g, '\u001D$1\u000F');                        // Apply italics
+}
+
+export function formatFromIRCToDiscord(text) {
+  // Regex doesn't work here, because formatting can be applied arbitrarily in IRC.
+  // Iterate over the string, and keep track of which formatting is active,
+  // and close all formatting if a reset symbol is encountered.
+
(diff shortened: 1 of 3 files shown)


ekmartin:
From a quick glance this looks really cool, thanks! Will do a more thorough review as fast as I have time :)

lib/formatting.js
```diff
@@ -0,0 +1,61 @@
+/* eslint-disable no-unused-vars */
+
+export function formatFromDiscordToIRC(text) {
+  // Apply formatting rules, from most specific to least specific.
+  return text.replace(/~~(.*?)~~/g, '$1')                        // Remove unsupported strikethrough
+    .replace(/__\*\*\*(.*?)\*\*\*__/g, '\u001F\u001D\u0002$1\u000F') // Apply underline-italics-bold
```
ekmartin: Maybe move `\u0001F` and similar to a constant? I.e.

```js
const ircFormatCharacters = {
  underline: '\u0001F',
  // and so on
};

File: lib/formatting.js [hunk 2]
```diff
@@ -0,0 +1,61 @@
...
+  };
+
+  let activeFormatters = [];
+  let resultString = '';
+
+  for (let i = 0; i < text.length; i++) {
+    const char = text[i];
+    if (char === '\u001D') { // italics
+      activeFormatters.push(FORMATTERS.ITALICS);
+      resultString = resultString.concat('*');
```
ekmartin: Wouldn't this be the same as `resultString += '*'`?

lib/formatting.js
```diff
@@ -0,0 +1,61 @@
...
+    } else if (char === '\u0002') { // bold
+      activeFormatters.push(FORMATTERS.BOLD);
+      resultString = resultString.concat('**');
+    } else if (char === '\u001F') { // underline
+      activeFormatters.push(FORMATTERS.UNDERLINE);
+      resultString = resultString.concat('__');
+    } else if (char === '\u000F') { // reset
+      // Reset all active formattings
+      for (let j = (activeFormatters.length - 1); j >= 0; j--) {
+        resultString = resultString.concat(FORMATTERS.discordSymbols[activeFormatters[j]]);
```
ekmartin: I might be misunderstanding something, but doesn't this need to happen at the end if no reset has been reached as well? I.e. `\u001FThis is underlined` needs to be turned into `__This is underlined__`?

lib/formatting.js
```diff
@@ -0,0 +1,61 @@
...
+    .replace(/\*(.*?)\*/g, '\u001D$1\u000F');                        // Apply italics
+}
+
+export function formatFromIRCToDiscord(text) {
+  // Regex doesn't work here, because formatting can be applied arbitrarily in IRC.
+  // Iterate over the string, and keep track of which formatting is active,
+  // and close all formatting in order if a reset symbol is encountered.
+
+  const FORMATTERS = {
+    BOLD: 1,
```
ekmartin: Wouldn't it work to just map directly to symbol right away? I.e. `BOLD: '**'`

test/formatting.test.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable prefer-arrow-callback */
+
+import chai from 'chai';
+import { formatFromDiscordToIRC, formatFromIRCToDiscord } from '../lib/formatting';
+
+const expect = chai.expect;
+
+describe('Formatting', function() {
+  it('should parse a discord string into an irc string', function() {
```
ekmartin: If you've got time I think the tests would be a lot more useful if split into multiple smaller ones, testing things like with or without an ending reset and similar.

Also, the rest of the tests use `should` - consistency is good :) 
If you want to you can use an arrow function here as well, since you don't use `this` inside the test cases.

Kurocon:
Thanks for the comments, you make good points. I'll try to address them as soon as I can. I don't have a lot of experience coding in Javascript, most of my time is spent in Python or Java, so I'm not really familiar with the specifics of the packages, especially the testing framework you're using. Nonetheless, I'll try to fix the issues soon!

ekmartin:
No problem, was just minor stuff :)

ekmartin:
Closing this due to inactivity. Feel free to reopen if you get time to fix the proposed changes!

Issue #63, opened 2016-04-30
Support for Per Server Nicknames


smashedr:
Discord has a new feature, that allows per server nicknames. When I set a server specific nickname, the bot still posts to IRC with my global nickname, not the nickname I had on the server that the bot is on.

ekmartin:
I'm not completely sure I follow you, are you talking about the nickname that is used when the bot is posting to IRC? Because this is decided by the config field `nickname`, and doesn't have anything to do with the bot's username on the Discord side. 

https://github.com/reactiflux/discord-irc#example-configuration

TheDoctorsLife:
I think he's referring to per server nicknames, as announced in this blog post. From the looks of it this would have to wait for discord.js to update with support before any work can go into this. https://blog.discordapp.com/2016-4-29-change-log/

malleusinferni:
It looks like discord.js has nickname support as of 8.0 (see #78).

alienth:
Support now exists, but discord-irc will have to have some changes to actually use the newly exposed info. API info is here: http://discordjs.readthedocs.io/en/indev/docs_server.html#detailsofuser-user

devalnor:
+1

omnicons:
+1

PR #107, opened 2016-08-24
resolves reactiflux/discord-irc#63 nicknames

DarkSpyro003:
Use per server nickname instead of username.
See issue reactiflux/discord-irc#63

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -137,10 +137,12 @@ class Bot {
     logger.debug('Channel Mapping', channelName, this.channelMapping[channelName]);
     if (ircChannel) {
       const username = author.username;
+      let nickname = author.username;
       const fromServer = message.server;
-      const userDetails = fromServer.detailsOfUser(author);
-      let nickname = userDetails.nick;
-      if (nickname == null) nickname = author.username;
+      if (fromServer != null) {
+        const userDetails = fromServer.detailsOfUser(author);
+        if (userDetails != null && userDetails.nick != null) nickname = userDetails.nick;
+      }
       let text = this.parseText(message);
       let displayUsername = nickname;
       if (this.ircNickColor) {
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -140,7 +140,7 @@ class Bot {
       const fromServer = message.server;
       const userDetails = fromServer.detailsOfUser(author);
       let nickname = userDetails.nick;
-      if(nickname == null) nickname = author.username;
+      if (nickname == null) nickname = author.username;
       let text = this.parseText(message);
       let displayUsername = nickname;
       if (this.ircNickColor) {
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -137,11 +137,15 @@ class Bot {
     logger.debug('Channel Mapping', channelName, this.channelMapping[channelName]);
     if (ircChannel) {
       const username = author.username;
+      const fromServer = message.server;
+      const userDetails = fromServer.detailsOfUser(author);
+      let nickname = userDetails.nick;
+      if(nickname == null) nickname = author.username;
       let text = this.parseText(message);
-      let displayUsername = username;
+      let displayUsername = nickname;
       if (this.ircNickColor) {
-        const colorIndex = (username.charCodeAt(0) + username.length) % NICK_COLORS.length;
-        displayUsername = irc.colors.wrap(NICK_COLORS[colorIndex], username);
+        const colorIndex = (nickname.charCodeAt(0) + nickname.length) % NICK_COLORS.length;
+        displayUsername = irc.colors.wrap(NICK_COLORS[colorIndex], nickname);
       }
 
       if (this.isCommandMessage(text)) {
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -187,8 +187,23 @@ class Bot {
       }
 
       const withMentions = text.replace(/@[^\s]+\b/g, match => {
-        // TODO DarkSpyro003: Mentions with nicknames? What about usernames & nicknames with spaces?
-        const user = this.discord.users.get('username', match.substring(1));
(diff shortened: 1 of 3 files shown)


DarkSpyro003:
Sorry, I forgot to run the test suite before creating the initial PR. It should work now.

lib/bot.js
```diff
@@ -137,11 +137,17 @@ class Bot {
     logger.debug('Channel Mapping', channelName, this.channelMapping[channelName]);
     if (ircChannel) {
       const username = author.username;
+      let nickname = author.username;
+      const fromServer = message.server;
+      if (fromServer != null) {
```
ekmartin: When would this be null/undefined?

lib/bot.js
```diff
@@ -137,11 +137,17 @@ class Bot {
...
+        const userDetails = fromServer.detailsOfUser(author);
+        if (userDetails != null && userDetails.nick != null) nickname = userDetails.nick;
```
ekmartin: could do `nickname = _.get(userDetails, 'nick', nickname)` if you want. See https://lodash.com/docs#get

lib/bot.js
```diff
@@ -137,10 +137,12 @@ class Bot {
     logger.debug('Channel Mapping', channelName, this.channelMapping[channelName]);
     if (ircChannel) {
       const username = author.username;
+      let nickname = author.username;
       const fromServer = message.server;
-      const userDetails = fromServer.detailsOfUser(author);
```
ekmartin: The code was a lot cleaner before this commit. The tests failed because the stub messages don't include the server property. If this is always present in real world situations it's better to change the tests to include it. The updates needed are in [this file](https://github.com/reactiflux/discord-irc/blob/master/test/bot.test.js#L73-L85), wherever `sendToIRC` is called.

You would need to create a stub server, so that you can stub out the `detailsOfUser` method. Since we don't need to pass a constructor anywhere, this is simpler than when stubbing out discord.js's client itself.
Here's an example of how it could look:

```js
const createServerStub = nickname => (
  sandbox.stub().resolves({ 
    detailsOfUser() {
      return { nick: nickname };
    }
  })
);

// Later, in the test itself:
const nickname = 'nick';
const server = createServerStub(nickname);
const message = {
  // other message stuff
  server
};

server.detailsOfUser.should.have.been.calledWith(nickname);
```

ekmartin:
Thanks for fixing this! Let me know if you more help with the test stuff :)

lib/bot.js
ekmartin: See my last comment, here: https://github.com/reactiflux/discord-irc/pull/107#discussion_r76358268

test/bot.test.js
```diff
@@ -39,6 +39,14 @@ describe('Bot', function() {
...
   });
 
+  const createServerStub = function(nickname) {
+    return {
+      detailsOfUser(username) {
+        return { nick: nickname };
+      }
+    };
+  };
+
```
DarkSpyro003: I couldn't get your example to work. Would something like this be fine?

DarkSpyro003:
the lodash _.get wouldn't work because the value would be explicitly null, not undefined.

lib/bot.js
```diff
@@ -137,11 +137,15 @@ class Bot {
     logger.debug('Channel Mapping', channelName, this.channelMapping[channelName]);
     if (ircChannel) {
       const username = author.username;
+      const fromServer = message.server;
+      const userDetails = fromServer.detailsOfUser(author);
+      let nickname = userDetails.nick;
```
ekmartin: These two could be changed to:
`const nickname = userDetails.nick || author.username;` I think?

test/bot.test.js
ekmartin: Maybe I wrote something wrong in the last one, here's how it would be with an arrow function at least:

```js
const createServerStub = (nickname = null) => ({
  detailsOfUser() {
    return { nick: nickname };
  }
});
```

Note the `nickname = null`, that way you wont have to explicitly call it with `null` since it defaults to it (you could do this with your code to as well, don't need an arrow function - it just looks a bit cleaner).

DarkSpyro003:
I expanded this PR to convert mentions using Discord server nicknames as well. ie IRC users will see mentions with the nickname now, and they can send @mentions with the server nickname now.

lib/bot.js
```diff
@@ -179,7 +188,23 @@ class Bot {
...
+        if (user) {
+          if (this.getDiscordNicknameOnServer(user, server).toUpperCase()
+          !== match.substring(1).toUpperCase()) {
+            user = null; // Nickname doesn't match mention
+          }
+        }
+        if (!user) {
+          const length = this.discord.users.length;
+          for (let i = 0; i < length && !user; i++) {
+            if (this.getDiscordNicknameOnServer(this.discord.users[i], server).toUpperCase()
```
ekmartin: Wouldn't this create an API call for every discord user on the server, and do that for every single person someone has mentioned?

ekmartin:
I've merged in the server nickname functionality that fixes #63, and [updated](https://github.com/reactiflux/discord-irc/commit/44849df01b6d9e7632ee32fa609c0e0eaa9b1e10) it to the new discord.js v9 APIs. It's released in version [2.1.0](https://github.com/reactiflux/discord-irc/blob/master/CHANGELOG.md#210---2016-10-09)  and works flawlessly - thanks! :)

In regards to the mention part, the API call issue I mentioned above seems to be wrong, as it looks like it retrieves all the guild info (including the members list) in one go. I think the code could need some refactoring to be slightly more readable though, but if you're still interested in fixing it feel free to open a new PR for that and we can discuss it there.

By the way, it's usually easiest to merge as small pull requests as possible - so if the first proposed change doesn't need the new functionality to work as intended it's usually better to split it into multiple PRs.

PR #134, opened 2016-10-07
Send text messages only to text discord channels

dustinlacewell:
This adds a function predicate for finding channels so that only text-channels are selected for text relay. Also adds a debug log in the case the select channel can't receive text messages.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -170,8 +170,10 @@ class Bot {
   sendToDiscord(author, channel, text) {
     const discordChannelName = this.invertedMapping[channel.toLowerCase()];
     if (discordChannelName) {
-      // #channel -> channel before retrieving:
-      const discordChannel = this.discord.channels.find('name', discordChannelName.slice(1));
+      // #channel -> channel before retrieving and select only text channels:
+      const discordChannel = this.discord.channels
+        .filter(c => c.type === 'text')
+        .find('name', discordChannelName.slice(1));
 
       if (!discordChannel) {
         logger.info('Tried to send a message to a channel the bot isn\'t in: ',


lib/bot.js
```diff
@@ -170,15 +170,22 @@ class Bot {
...
+        (o) => o.name === discordChannelName.slice(1) && o.type === 'text'
+      );
 
       if (!discordChannel) {
         logger.info('Tried to send a message to a channel the bot isn\'t in: ',
           discordChannelName);
         return;
       }
 
+      if (typeof discordChannel.sendMessage !== 'function') {
```
ekmartin: This will never happen now though right? Since you've filtered out the channels that aren't text.

lib/bot.js
```diff
@@ -170,15 +170,22 @@ class Bot {
   sendToDiscord(author, channel, text) {
     const discordChannelName = this.invertedMapping[channel.toLowerCase()];
     if (discordChannelName) {
-      // #channel -> channel before retrieving:
-      const discordChannel = this.discord.channels.find('name', discordChannelName.slice(1));
+      // #channel -> channel before retrieving and select only text channels:
+      const discordChannel = this.discord.channels.find(
+        (o) => o.name === discordChannelName.slice(1) && o.type === 'text'
```
ekmartin: You'll have to remove the parenthesis around `(o)` to pass the linter.
Might be cleaner to write it something like this though, maybe?
```js
this.discord.channels
  .filter(c => c.type === 'text')
  .find('name', discordChannelName.slice(1));
```

dustinlacewell: Apparently `filter` is not a function on the discordChannel?

ekmartin: Oh, sorry - I see what you mean now.

ekmartin: You have to update the stub here: https://github.com/reactiflux/discord-irc/blob/master/test/stubs/discord-stub.js#L13

Since our tests don't actually talk to discord we stub out that part in the tests.
You can stub this out like this:
```js

this.channels = {
  filter() {
    return this.channels;
  },
  get: this.getChannel,
  find: this.getChannel
};

ekmartin: So basically just add the filter() function to the object I linked you.

ekmartin: @dustinlacewell if you don't have time to fix the tests I can do that for you when merging this. Let me know what you think is best :)

ekmartin:
This has been released in version [2.1.3](https://github.com/reactiflux/discord-irc/blob/master/CHANGELOG.md#213---2016-11-02) - thanks! (https://github.com/reactiflux/discord-irc/commit/c3881be0bbf48aea62bc83bff1ab0231b00b838d)

Issue #137, opened 2016-10-10
Mentioning by nickname


altyr:
Thanks for making the bot display discord nicknames in IRC.
Currently _@_nickname doesn't work from IRC, are there plans to allow mentioning people by their nicknames? If not I could take a stab at this.

ekmartin:
@DarkSpyro003 fixed this in https://github.com/reactiflux/discord-irc/pull/107 - but there were still a few issues to resolve so I urged her to open a new PR. If she doesn't have time I'm sure she wouldn't mind if you finished what she started, but should give it a little time first :)

DarkSpyro003:
The basic functional code for it is in my fork (in branch server-nicknames DarkSpyro003/discord-irc:server-nicknames ) but it still needs some refactoring. There's probably a better way to write it than I did.
I don't have a lot of time for this and I generally don't write JavaScript.
If you want you can take a look at it.

DarkSpyro003:
I've picked this issue up again and have made changes to make it work with the current API version.
I still have to fix the tests and do a thorough test and hope to have a pull request ready tomorrow.

WIP on develop branch https://github.com/DarkSpyro003/discord-irc/tree/server-nicknames-develop

ekmartin:
Great @DarkSpyro003 - looking forward to the pull request!

altyr:
Deployed @DarkSpyro003's version to test, looks good so far. Nice work.
I wonder if the regex used to match the name could be sensibly expanded to catch nicknames with spaces in them? On the other hand that might create a stack of other problems...

DarkSpyro003:
I've thought about that as well @altyr but I think the only way to do that would have serious performance rammifications. You'd probably need a list of all the nicknames in the discord channel, and then search for occurences of @<nickname> of each of those to determine if someone was trying to mention them.
Not to mention, what would happen if someone was called <abc> and someone else <abc def> and someone wrote @abc def ?
Lets say it appears to be non-trivial and beyond the scope of this feature.

ekmartin:
Published in version [2.1.4](https://github.com/reactiflux/discord-irc/blob/master/CHANGELOG.md#214---2016-11-14) - thanks again @DarkSpyro003.

dumptruckman:
Do mentions coming from irc only work with the at sign? Traditionally, an IRC client will ping you if someone merely types your name. Is it suppose to work that way?

altyr:
[...]

Issue #146, opened 2016-11-06
Crashing when Discord Webhook messages are sent to the channel


Psykus0:
Hi,

I'm using the webhook feature of Discord to send messages from IFTTT to a Discord channel, and whenever they are sent, discord-irc crashes and the message does not display in the IRC channel before discord-irc disconnects.

https://support.discordapp.com/hc/en-us/articles/228383668-Intro-to-Webhooks

I got the exact error message below:
```js
/usr/lib/node_modules/discord-irc/dist/bot.js:156
      const nickname = userDetails.nickname || author.username;
                                  ^

TypeError: Cannot read property 'nickname' of null
    at Bot.sendToIRC (/usr/lib/node_modules/discord-irc/dist/bot.js:156:35)
    at Client.discord.on.message (/usr/lib/node_modules/discord-irc/dist/bot.js:108:12)
    at emitOne (events.js:96:13)
    at Client.emit (events.js:188:7)
    at MessageCreateHandler.handle (/usr/lib/node_modules/discord-irc/node_modules/discord.js/src/client/websocket/packets/handlers/MessageCreate.js:9:34)
    at WebSocketPacketManager.handle (/usr/lib/node_modules/discord-irc/node_modules/discord.js/src/client/websocket/packets/WebSocketPacketManager.js:93:65)
    at WebSocketManager.eventMessage (/usr/lib/node_modules/discord-irc/node_modules/discord.js/src/client/websocket/WebSocketManager.js:198:31)
    at WebSocket.ws.onmessage (/usr/lib/node_modules/discord-irc/node_modules/discord.js/src/client/websocket/WebSocketManager.js:74:37)
    at WebSocket.onMessage (/usr/lib/node_modules/discord-irc/node_modules/ws/lib/WebSocket.js:442:14)
    at emitTwo (events.js:106:13)
```

I installed discord-irc through npm and am on version 3.10.8 according to the output of "npm -v discord-irc"

Let me know if you need any info or help reproducing the issue and I'd be glad to help.

ekmartin:
I guess there's two options here, either filter them out or send them to IRC with the webhook's name as username (i.e. `<GitHub> New commit to...`). 

I think I like the last option best, but not sure. What do you think?

Psykus0:
Yeah the last option sounds great.

Psykus0:
This seems fixed in the newest version on GitHub :+1:

ekmartin:
Cool, then it got fixed in #147.

PR #147, opened 2016-11-09
Mentioning by guild nickname

DarkSpyro003:
resolves issue #137

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -137,10 +137,12 @@ class Bot {
     logger.debug('Channel Mapping', channelName, this.channelMapping[channelName]);
     if (ircChannel) {
       const username = author.username;
+      let nickname = author.username;
       const fromServer = message.server;
-      const userDetails = fromServer.detailsOfUser(author);
-      let nickname = userDetails.nick;
-      if (nickname == null) nickname = author.username;
+      if (fromServer != null) {
+        const userDetails = fromServer.detailsOfUser(author);
+        if (userDetails != null && userDetails.nick != null) nickname = userDetails.nick;
+      }
       let text = this.parseText(message);
       let displayUsername = nickname;
       if (this.ircNickColor) {
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -140,7 +140,7 @@ class Bot {
       const fromServer = message.server;
       const userDetails = fromServer.detailsOfUser(author);
       let nickname = userDetails.nick;
-      if(nickname == null) nickname = author.username;
+      if (nickname == null) nickname = author.username;
       let text = this.parseText(message);
       let displayUsername = nickname;
       if (this.ircNickColor) {
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -137,11 +137,15 @@ class Bot {
     logger.debug('Channel Mapping', channelName, this.channelMapping[channelName]);
     if (ircChannel) {
       const username = author.username;
+      const fromServer = message.server;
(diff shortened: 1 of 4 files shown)


lib/bot.js
```diff
@@ -108,11 +108,21 @@ class Bot {
     });
   }
 
+  static getDiscordNicknameOnServer(user, guild) {
+    const userDetails = guild.members.find('id', user.id);
+    if (userDetails) {
+      return userDetails.nickname || user.username;
+    }
+    return user.username;
```
DarkSpyro003: I'll check if this is actually ever reachable.

DarkSpyro003: This is reached when sent by a webhook. Looks like it resolves #146

lib/bot.js
```diff
@@ -183,7 +193,19 @@ class Bot {
...
-        const user = this.discord.users.find('username', match.substring(1));
+        const search = match.substring(1);
+        const guild = discordChannel.guild;
+        let user = guild.members.find('nickname', search);
+        if (!user) {
+          user = this.discord.users.find('username', search);
+          if (user) {
+            const serverUser = guild.members.find('id', user.id);
+            if (serverUser.nickname && serverUser.nickname !== search) {
+              // User has a guild nickname which is different from the search query
```
ekmartin: Not sure, but maybe it wouldn't be a terrible idea to just let people be highlighted by usernames as well?
Nonetheless, I think this could be refactored slightly by returning early instead. Example:
```js
const nickUser = guild.members.find('nickname', search);
if (nickUser) {
  return nickUser;
}

const user = this.discord.users.find('username', search);
if (user && guild.members.find('id').nickname === search) {
  return user;
}

return match;
```

test/bot.test.js
```diff
@@ -378,4 +385,32 @@ describe('Bot', function () {
...
+    const text = 'Hello, @somenickname!';
+    const expected = `**<${username}>** Hello, <@${testUser.id}>!`;
+
+    this.bot.sendToDiscord(username, '#irc', text);
+    this.sendMessageStub.should.have.been.calledWith(expected);
+  });
+
+  it('should not convert username mentions from IRC if nickname differs', function () {
+    const testUser = new discord.User(this.bot.discord, { username: 'testuser', id: '123', nickname: 'somenickname' });
+    this.findUserStub.withArgs('username', testUser.username).returns(testUser);
```
ekmartin: nice sinon usage!

lib/bot.js
ekmartin: Great!

lib/bot.js
DarkSpyro003: @ekmartin I'm not sure that's a good idea. The people on IRC would be unable to see what the usernames of users on Discord are, and could possibly even inadvertently mention someone they weren't trying to mention.

DarkSpyro003: I had to change this refactor around a bit so it still works for people without nicknames.

ekmartin: Wouldn't `guild.members.find('id').nickname` just be null/undefined for users without nicknames? If so you'd just be comparing `search` to that anyway, which would never be true since search is the match from your regex - right?

DarkSpyro003: @ekmartin If user found by username and nickname = null it means the "visible nickname" is a match and should return true.

ekmartin:
I've added you as a collaborator @DarkSpyro003 - thanks for the help :)

PR #166, opened 2016-11-30
Strip IRC formatting before sending to Discord

rossengeorgiev:


ekmartin:
Maybe a better idea would be to convert the formatting to their Discord counter part, and vice versa? This was implemented by @Kurocon in #97, but the PR was never completed. It's a good implementation though, just needed some finishing touches. If you have time I'm sure he wouldn't mind if you extended his PR and fixed the relevant feedback.

rossengeorgiev:
I can't remember seeing anyone ever use underline, italic or bold on IRC for all the years. It is a nice idea, but I don't see any use for it. I am not interested in working on that PR.  Just want clean readable text.

If that PR is ever gets touched, it can easily remove the 2 lines from this one.

ekmartin:
That's alright, I could agree with merging this in the meantime. Maybe use an existing library for it instead though? E.g. [irc-colors.js](https://github.com/fent/irc-colors.js) has a `stripColorsAndStyle`function (which seems to have tests as well).

rossengeorgiev:
Why add a dependency for something so simple?

ekmartin:
@rossengeorgiev: Because it adds tests for that specific function as well. Regarding the small module discussion there's plenty of arguments here: https://github.com/sindresorhus/ama/issues/10#issuecomment-117766328

I don't really think this is the same discussion as that though, as the function you're adding here really isn't that simple - if you're not overly familiar with IRC formatting escape sequences. That's fine though, _as long as it's tested_, which in your case it's not.

StrikerMan780:
"I don't see any use for it."

I can. I run an IRC bot that gets the output of a game server's chat, and it color-codes the player names based on their team. It also converts the game's chat color codes to IRC ones.

ekmartin:
@StrikerMan780: I'm open to both implementations, but I'm going to close this PR due to inactivity. If anyone wants to give it a go, feel free to read the comments above and send a new PR.

PR #187, opened 2017-02-04
Support using discord channel ids

mraof:


ekmartin:
Would you care to explain why you think is useful as well? Can't see any immediate gain, but I might be missing something.

mraof:
It allows the config to keep working even if discord channels are renamed, which is the problem I was having.
It might also allow specifying which channel if the bot is in multiple servers that have a channel named the same, but I haven't tested that

lib/bot.js
```diff
@@ -193,9 +194,9 @@ class Bot {
     const discordChannelName = this.invertedMapping[channel.toLowerCase()];
     if (discordChannelName) {
       // #channel -> channel before retrieving and select only text channels:
-      const discordChannel = this.discord.channels
+      const discordChannel = discordChannelName.charAt(0) === '#' ? this.discord.channels
```
ekmartin: `discordChannelName.startsWith('#')`

lib/bot.js
```diff
@@ -154,7 +154,8 @@ class Bot {
     if (author.id === this.discord.user.id) return;
 
     const channelName = `#${message.channel.name}`;
-    const ircChannel = this.channelMapping[channelName];
+    const ircChannel = this.channelMapping[message.channel.id in this.channelMapping ?
```
ekmartin: `const ircChannel = this.channelMapping[message.channel.id] || this.channelMapping[channelName]` maybe?

ekmartin:
Ye, that makes sense. Care to add a test or two?

mraof:
I'm afraid I don't actually know how to add tests

Sanqui:
I just ran an issue when joining the bot onto two Discord servers with #﻿general.  Would appreciate this a lot.  Might give writing tests a shot but I'm not familiar with it at all.

Throne3d:
I've had a go at adding tests for this in PR #202. Thanks for putting in the work for the functionality.

ekmartin:
Closing this since #202 was merged. Thanks!

PR #193, opened 2017-02-26
Avoid self-message highlights on IRC

Sanqui:
When a user connected to both Discord and IRC speaks on Discord, they will likely get unwanted highlights in their IRC client.  This is a fix which I believe is standard practice.  By putting a zero width space character between each letter of the display name, we ensure IRC clients do not detect the name and highlight on it.

The other line in the commit is a fix where the wrong variable was being used.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -162,9 +162,14 @@ class Bot {
       const nickname = Bot.getDiscordNicknameOnServer(author, fromGuild);
       let text = this.parseText(message);
       let displayUsername = nickname;
+
+      // Join the display name with double "bold" formatting characters
+      // to prevent unintentional highlights on the IRC side
+      displayUsername = displayUsername.split('').join('\u200d');
+
       if (this.ircNickColor) {
         const colorIndex = (nickname.charCodeAt(0) + nickname.length) % NICK_COLORS.length;
-        displayUsername = irc.colors.wrap(NICK_COLORS[colorIndex], nickname);
+        displayUsername = irc.colors.wrap(NICK_COLORS[colorIndex], displayUsername);
       }
 
       if (this.isCommandMessage(text)) {
--- a/test/bot.test.js
+++ b/test/bot.test.js
@@ -98,7 +98,7 @@ describe('Bot', function () {
     };
 
     bot.sendToIRC(message);
-    const expected = `<${message.author.username}> ${text}`;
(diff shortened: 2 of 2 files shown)


Sanqui:
Oops, I can see the tests are failing.  I couldn't get them working on my side, so I was hoping they're not particularly extensive :)  I'll see about fixing them.

Sanqui:
I have fixed the tests, but you may not be very happy about how they look.  :)
Let me know what you think.

Sanqui:
Hm, somebody has already reported having display issues where the non-breaking zero width characters (U+FEFF) show up as spaces for them.  However, breaking ones (U+200B) do work.  So I suppose this probably requires more testing across clients.

Sanqui:
I have cleaned this pull request up.  It now uses zero width space characters which seem to work in all clients so far.

lib/bot.js
```diff
@@ -162,9 +162,14 @@ class Bot {
       const nickname = Bot.getDiscordNicknameOnServer(author, fromGuild);
       let text = this.parseText(message);
       let displayUsername = nickname;
+
+      // Join the display name with double "bold" formatting characters
+      // to prevent unintentional highlights on the IRC side
+      displayUsername = displayUsername.split('').join('\u200d');
```
ekmartin: Maybe merge this with line 164? I.e. `let displayUsername = nickname.split('').join('\u200d');`

test/bot.test.js
```diff
@@ -98,7 +98,7 @@ describe('Bot', function () {
     };
 
     bot.sendToIRC(message);
-    const expected = `<${message.author.username}> ${text}`;
+    const expected = `<${message.author.username.split('').join('\u200d')}> ${text}`;
```
ekmartin: could add a simple helper function somewhere towards the top of the file that does the splitting?
i.e.

```js
const escapeUsername = username => username.split('').join('\u200d');
```

lib/bot.js
Sanqui: I think this is more explicit since it doesn't combine the operation with getting the name from another variable, but I can do it.

ekmartin:
Tested with Weechat, seems to be working well. Thanks! Released in version [2.2.0](https://github.com/reactiflux/discord-irc/blob/master/CHANGELOG.md#220---2017-03-06).

PR #194, opened 2017-03-05
Add the option to prevent the user who uses both Discord and IRC from being notified by messages they sent themself

Kroisse:
It's done by inserting [U+2060 word joiner][1] between the Discord nickname in the message that sent to IRC.

[runes][] was added to the dependencies to handle non-ASCII username properly, like this:

Some unicode "characters" like รื or 👬🏻 is consist of more than 1 codepoints, so [String.substr()][] could break them.

[1]: http://www.fileformat.info/info/unicode/char/2060/index.htm
[runes]: https://www.npmjs.com/package/runes
[String.substr()]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,11 @@ First you need to create a Discord bot user, which you can do by following the i
     "ircNickColor": false, // Gives usernames a color in IRC for better readability (on by default)
     // Makes the bot hide the username prefix for messages that start
     // with one of these characters (commands):
-    "commandCharacters": ["!", "."]
+    "commandCharacters": ["!", "."],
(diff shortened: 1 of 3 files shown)


ekmartin:
Hi! @Sanqui's #193 also fixed this. However, I just realized that it broke longer unicode characters, like the issue you describes. I think a good solution would be to use runes' split function here: https://github.com/reactiflux/discord-irc/blob/master/lib/bot.js#L168

Would you want to send another PR fixing that? I don't think we need a configuration option for this, as I don't think it breaks anyone's existing user flow.

PR #202, opened 2017-03-25
Add Discord channel ID matching (PR #187) and tests for it

Throne3d:
Builds on PR #187 (credit to @mraof for adding in the actual change to the code). This adds tests, where I think they're supposed to be? It doesn't actually test that Discord – or the library you're using to interface with it – supports IDs in its API, but this seems to work when I use them in my own config, and these tests should ensure the bot's code itself has no trouble with the IDs. 

(IDs can be retrieved by writing, e.g., `\#general`, which should produce something of the form `<#00000000>`. Remove the `<#` and `>` and use that in the config.)

Commit 2ae9e37 was added since the previous check seems to break with this new addition to the config file – it seems to be related to how the items are ordered in the hash and so how they get added to the `channels` list, which I'm pretty sure isn't *important* anywhere, so I modified it to a check for presence instead of a check that the first item in the list is precisely that.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -154,7 +154,7 @@ class Bot {
     if (author.id === this.discord.user.id) return;
 
     const channelName = `#${message.channel.name}`;
-    const ircChannel = this.channelMapping[channelName];
+    const ircChannel = this.channelMapping[message.channel.id in this.channelMapping ? message.channel.id : channelName];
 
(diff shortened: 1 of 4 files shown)


ekmartin:
Awesome! Thanks @mraof and @Throne3d.

ekmartin:

Issue #186, opened 2017-01-30
Role mensions display as number on irc


PieterHollevoet:
When you @ mention a role on discord, It displays on IRC as numbers. 
For example
discord : @sysadmin  ==> irc : @&233558485999...

Could you give me some pointers to where the problem lies and we will fix it ourselves.

ekmartin:
`parseText` is responsible for translating user and channel highlights to a readable format.
Something like the one for channels should do the trick: https://github.com/reactiflux/discord-irc/blob/master/lib/bot.js#L140-L143

Pull request welcome!

Throne3d:
Fixed in #203.

PR #203, opened 2017-03-25
Parse role mentions appropriately, as with channel and user mentions

Throne3d:
## Description
As requested in #186, translate role mentions appropriately between IRC and Discord. This works both ways, so the numbers will be translated into proper text when outputted in IRC, and the plain-text mentions will be made into proper role mentions when outputted in Discord, if (and only if) the role is mentionable.

## Present behavior
As in #186, using e.g. `@sysadmin` in Discord will end up outputting a long string into IRC, e.g. `<@&233558485999…>`. In addition, using `@sysadmin` in IRC will not produce a mention in Discord, even if the sysadmin role is mentionable.

## Modified behavior
If `newrole` is not mentionable and `otherrole` is mentionable, the following output is produced in Discord:
```
Throne3d wrote:
@newrole

IrcBridge wrote:
<Throne> @newrole
<Throne> @otherrole # actually works

Throne3d wrote:
@otherrole # actually works
```
and in IRC:
```
<test2> <Throne3d> @newrole
<Throne> @newrole
<Throne> @otherrole
<test2> <Throne3d> @otherrole
```

I wasn't sure how exactly to set up the tests (you may note the first commit, a1dfdf6, sets up a `const getRole` in `bot.test.js`, which is later removed in favor of a `this.findRoleStub`, behaving somewhat like `this.findUserStub`), but they seem to pass where appropriate and match the style of the other tests at least reasonably well.

This should solve #186.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -141,6 +141,11 @@ class Bot {
         const channel = this.discord.channels.get(channelId);
         return `#${channel.name}`;
       })
+      .replace(/<@&(\d+)>/g, (match, roleId) => {
+        const role = message.guild.roles.get(roleId);
+        if (role) return `@${role.name}`;
+        return '@deleted-role';
+      })
       .replace(/<(:\w+:)\d+>/g, (match, emoteName) => emoteName);
   }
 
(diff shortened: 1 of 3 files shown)


ekmartin:
Beautiful pull request - thank you!

ekmartin:

Issue #13, opened 2015-11-18
Custom formatting between Discord and IRC


ProfX66:
It would be really cool if I could customize the format that shows between the two services.

Like, I would like to strip the nick color that is sent to IRC, or have it show as [NICK] instead of <NICK>.

It would be really cool if it could just pull the rank color from Discord and use that instead of a random color.

FastiJan:
I have similar issue, in quakenet channel mode +c prevents the usage of control codes(colors/bolds/etc). The bot cannot send to channel, because it forces the nick to have color. I tried using ircOptions stripColors, but it doesn't seem to work.

FastiJan:
Got it somewhat working with no colors by just changing a line in
lib/bot.js Line ~147:

text = `<${coloredUsername}> ${text}`;
to
text = `<${username}> ${text}`;

then run build again.

unsure what else this might do or not do to work with channel flag +c.

ekmartin:
@FastiJan, version 0.7.0 now includes an option that lets you disable irc colors by setting `ircNickColor` to false.

PR #204, opened 2017-03-25
Add custom formatting to IRC & Discord output

Throne3d:
As requested by #13, it might sometimes be useful to have custom formatting in the output for IRC and Discord.

I'm not sure how exactly to add tests for this, as it seems like it's going to need a separate configuration for (almost) each test, and this doesn't look like it'll fit into the current `bot.test.js` file very easily. From a brief glance I can't see a 'context' feature in Chai as there is in RSpec, which would allow me to limit the `beforeEach` to just the tests present, and add separate ones at the end for this. It's possible I'm missing something.

I'm also not too strongly attached to this idea and it might make it harder to implement some things in the future, so if it'd be preferable to drop this, I don't care too strongly. (If people want it desperately enough, they can always just patch these commits on – hopefully it's clearly documenting enough, and it doesn't seem to break current tests.)

I'd appreciate feedback on how to proceed. Thanks.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -177,10 +177,7 @@ class Bot {
   }
 
   static substitutePattern(message, patternMapping) {
-    return message.replace(patternMatch, (match, varName) => {
-      if (varName in patternMapping) return patternMapping[varName];
-      return match;
-    });
+    return message.replace(patternMatch, (match, varName) => patternMapping[varName] || match);
   }
 
   sendToIRC(message) {
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -9,6 +9,7 @@ import { formatFromDiscordToIRC, formatFromIRCToDiscord } from './formatting';
 const REQUIRED_FIELDS = ['server', 'nickname', 'channelMapping', 'discordToken'];
 const NICK_COLORS = ['light_blue', 'dark_blue', 'light_red', 'dark_red', 'light_green',
   'dark_green', 'magenta', 'light_magenta', 'orange', 'yellow', 'cyan', 'light_cyan'];
+const patternMatch = /{\$(.+?)}/g;
 
 /**
  * An IRC bot, works as a middleman for all communication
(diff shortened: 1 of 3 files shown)


Throne3d:
I guess I could add a new "describe" context, within the current one, for the tests that use a regular-config bot, and separate out the `beforeEach` into two bits? (Then use another context for new tests that don't.)

Throne3d:
I've added tests for this (which should hopefully now pass, making this not a WIP), but perhaps instead of creating a new `describe` block to separate out the default-config bot tests, I should just do as some tests have already and redefine `bot` with a different config in the tests that need it.

Opinions?

Throne3d:
Nm, undid the contexting and just manually used `const bot` in the appropriate tests.

lib/bot.js
```diff
@@ -148,6 +168,13 @@ class Bot {
     return this.commandCharacters.indexOf(message[0]) !== -1;
   }
 
+  static substitutePattern(message, patternMapping) {
+    return message.replace(patternMatch, (match, varName) => {
+      if (varName in patternMapping) return patternMapping[varName];
```
ekmartin: Could simplify this to `(match, varName) => patternMapping[varName] || match`?

lib/bot.js
```diff
@@ -33,6 +34,25 @@ class Bot {
     this.ircNickColor = options.ircNickColor !== false; // default to true
     this.channels = _.values(options.channelMapping);
 
+    this.formatCommandPrelude = options.formatCommandPrelude || 'Command sent from Discord by {$nickname}:';
+    this.formatIRCText = options.formatIRCText || '<{$displayUsername}> {$text}';
+    this.formatURLAttachment = options.formatURLAttachment || '<{$displayUsername}> {$attachmentURL}';
+    // "{$keyName}" => "variableValue"
```
ekmartin: Nitpick, but I think it makes more sense to put these comments above the variables.

lib/bot.js
```diff
@@ -33,6 +34,25 @@ class Bot {
     this.ircNickColor = options.ircNickColor !== false; // default to true
     this.channels = _.values(options.channelMapping);
 
+    this.formatCommandPrelude = options.formatCommandPrelude || 'Command sent from Discord by {$nickname}:';
```
ekmartin: Maybe create a `format` object and have these as properties? I.e.
```js
"format": {
  "commandPrelude": "...",
  "ircText": "...",
  "urlAttachment": "..."
}
```

or something like that? Then just add a `this.format = options.format || {};` above?

ekmartin:
Looks great, thanks! I've normally been against adding too many configuration options, but I think this could be useful.

lib/bot.js
Throne3d: The main reason I didn't was in case the patternMapping was a blank string. I guess this probably isn't going to happen anywhere (unless there was somehow a zero-length discord message? But there's a guard for that), so it should be fine to simplify.

lib/bot.js
Throne3d: Makes sense.

Throne3d:
(Rebased onto master, rebased to move comments, updated to work with the new formatting-bridging thing (`{$text}` is now relevantly formatted), added a commit to move the formatting configs to a `format` object, added a commit to simplify substitutePattern.)

ekmartin:

PR #205, opened 2017-03-25
Convert text styles between IRC/Discord

### LINKED ISSUE #195: Strip mIRC formatting characters
Messages from the IRC side with colors show up as numbers on the Discord side.

Could probably just use this: https://www.npmjs.com/package/irc-colors

rahatarmanahmed:
This is a continuation of #97, but instead of manually parsing the text ourselves, I've opted to use the `irc-formatter`, `irc-colors`, and `simple-markdown` packages to keep this project's code clean. (I think `simple-markdown` is the parser Discord uses but I assume that only because they have their own fork of it on github.)

image
image

Resolves #195

### DIFF
--- a/lib/formatting.js
+++ b/lib/formatting.js
@@ -55,7 +55,7 @@ export function formatFromIRCToDiscord(text) {
   }
 
   // Remove all color markers from the message, as discord does not support colored text
-  resultString = resultString.replace(/\u0003(\d\d(,\d\d)?)?/g, '');
+  resultString = resultString.replace(/\u0003(\d\d?(,\d\d?)?)?/g, '');
 
   return resultString;
 }
--- a/lib/formatting.js
+++ b/lib/formatting.js
@@ -18,12 +18,30 @@ export function formatFromDiscordToIRC(text) {
 }
 
 export function formatFromIRCToDiscord(text) {
-  const blocks = ircFormatting.parse(text);
-  return blocks.map((block) => {
-    let mdText = block.text;
-    if (block.italic || block.reverse) mdText = `*${mdText}*`;
-    if (block.bold) mdText = `**${mdText}**`;
-    if (block.underline) mdText = `__${mdText}__`;
-    return mdText;
-  }).join('');
+  const blocks = ircFormatting.parse(text).map(block =>
+    // Consider reverse as italic, some IRC clients use that
+    _.assign({}, block, { italic: block.italic || block.reverse })
+  );
+  let mdText = '';
+
+  for (let i = 0; i <= blocks.length; i += 1) {
+    // Default to unstyled blocks when index out of range
+    const block = blocks[i] || {};
+    const prevBlock = blocks[i - 1] || {};
+
+    // Add start markers when style turns from false to true
+    if (!prevBlock.italic && block.italic) mdText += '*';
+    if (!prevBlock.bold && block.bold) mdText += '**';
+    if (!prevBlock.underline && block.underline) mdText += '__';
+
+    // Add end markers when style turns from true to false
+    // (and apply in reverse order to maintain nesting)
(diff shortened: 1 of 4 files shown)


rahatarmanahmed:
Ah hell, there's some inconsistent conversion from irc->discord if you do some nested styles. `irc-formatter` isn't really an adequate package since it gives you a flat list of blocks. I'll probably try to write a package to build like a parse tree of irc message formats to fix this.

rahatarmanahmed:
Nevermind! Took a different approach to IRC->discord formatting with the same package that should have less weird results.

Throne3d:
I don't know if this is relevant, but I was poking around at doing a configuration to allow stripping of mIRC formatting and I *think* I found a bug in how `irc-colors.js` handles `stripColorsAndStyle` (see [here](https://github.com/fent/irc-colors.js/issues/12)). In case it's useful or you notice some weird behavior when poking around at things.

rahatarmanahmed:
Good to know. This PR only uses `irc-colors` for avoiding hardcoding the style character codes so it  shouldn't be affected.

But the funny coincidence is that I was the one who added the stripping functions in `irc-colors` 😮

lib/formatting.js
```diff
@@ -0,0 +1,47 @@
+import ircFormatting from 'irc-formatting';
+import SimpleMarkdown from 'simple-markdown';
+import colors from 'irc-colors';
+import _ from 'lodash';
+
+function mdNodeToIRC(node) {
+  let content = node.content;
+  if (_.isArray(content)) content = content.map(mdNodeToIRC).join('');
```
ekmartin: Could just use `Array.isArray` I guess?

lib/formatting.js
```diff
@@ -0,0 +1,47 @@
...
+
+export function formatFromDiscordToIRC(text) {
+  const markdownAST = SimpleMarkdown.defaultInlineParse(text);
+  return markdownAST.map(mdNodeToIRC).join('');
+}
+
+export function formatFromIRCToDiscord(text) {
+  const blocks = ircFormatting.parse(text).map(block =>
+    // Consider reverse as italic, some IRC clients use that
+    _.assign({}, block, { italic: block.italic || block.reverse })
```
ekmartin: The package has a Node >6 requirement anyway, since discord.js enforces that, so could do
```js
{
  ...block,
  { italic: block.italic || block.reverse }
}
```

ekmartin:
Thanks! This seems useful, and the implementation looks clean.

rahatarmanahmed:
Ah my bad, not sure how I overlooked the babel config and node requirements.

ekmartin:
I'll cut a release when we've got the last two outstanding PRs merged.
I've added you as a collaborator - feel free to help review PRs if you want to and have time :)

ekmartin:

StrikerMan780:
Got a question about this, is it possible to have IRC colors show in Discord, or is it only formatting like bold and underline?

rahatarmanahmed:
Only formatting, discord markdown doesn't have any coloring option (except for syntax highlighting)

lib/formatting.js
```diff
@@ -0,0 +1,47 @@
...
+  const blocks = ircFormatting.parse(text).map(block => ({
+    // Consider reverse as italic, some IRC clients use that
+    ...block,
+    italic: block.italic || block.reverse
+  }));
+  let mdText = '';
+
+  for (let i = 0; i <= blocks.length; i += 1) {
+    // Default to unstyled blocks when index out of range
+    const block = blocks[i] || {};
```
qaisjp, line 29: When can `blocks[i]` ever be `undefined`?

qaisjp: Ah, `i <= blocks.length` not `i < blocks.length`. Sneaky code.

PR #40, opened 2016-03-10
IRC Users joining/parting/quiting 

Ratismal:
I love Discord-IRC, but one of my major gripes with it is that IRC join/part/quit events aren't outputted to discord. I've just been manually changing my local files, but it's a bit irritating to do whenever I update. It would be great if this functionality came installed!

I have implemented this functionality in this branch. Feel free to change it as needed.

As a sidenote, I love the refactor that this code has received! It looks much cleaner.

Issue #92, opened 2016-08-01
show who is on IRC from discord and vice versa


ask-compu:
you should add commands so people on discord can see who is on IRC and vice versa, it would also be nice to see IRC nick changes, joins, quits, and parts from discord, maybe make it an option to show or hide such messages

creesch:
You can do the other way around easily by enabling discord widgets. This would allow you to have a window open with the people active on discord on it. If you use a webbased irc client like irccloud you can even make a userscript and append it to the userlist.

ask-compu:
that is not ideal and doesnt address showing whos on irc from the discord side

creesch:
I realize that, I was just giving a suggestion that works until such commands as you asked for are implemented.

dwhagar:
Showing who is on the IRC side can be done via a /names command, the response from that shows all nicks in the channel.  A subsequent /who commands on each nick can give more information, but the /names should provide all the information one would need.

ask-compu:
@dwhagar but u cant do /names and /who through the bridge from discord

dwhagar:
Yes but the bridge could relay the commands and process the responses.

ask-compu:
@dwhagar that could work but it might be best to give the response in PM on either end so it doesnt spam the channel/server

dwhagar:
I agree I have an IRC bot written in python3 that processes names and who responses.

ask-compu:
@dwhagar i have a python3 irc bot that does all kinds of stuff

dwhagar:
Nice we should compare notes.

ask-compu:
@dwhagar just click on my profile and look for the compubot repository

LordAlderaan:
I got it working one way. Discord is showing a list of people in IRC in the subject of a channel by changing bot.js in the following matter.

Add to the attachListeners section of the Bot class:
```javascript
	// When a list of names on a channel is received on IRC set the topic on the corresponding Discord channel to a list of those names
    this.ircClient.on('names', (channel, nicks) => {
      _winston2.default.debug('Names:', channel, ' Nicks:', Object.keys(nicks).join(', '));
      this.setTopicDiscord(channel,`Online at IRC: ${Object.keys(nicks).join(', ')}`);
    });
    
	// Trigger requesting a list of names on a IRC channel
    this.ircClient.on('join', (channel, nick, message) => {
      if (nick !== this.nickname) {
        _winston2.default.debug('Join:', channel, nick);
[...]

PR #207, opened 2017-03-26
Bridge IRC join/part/quit messages to Discord

Throne3d:
As in PR #40, this bridges the join/part/quit messages from IRC to Discord. It builds on that PR (thanks, @Ratismal), rebasing it onto master (and accidentally doing something stupid in the rebase, which a later commit fixes).

This adds two config settings:

* ~~`disableJoinPartMsgs` – if set to true, the bot won't announce any~~ `ircStatusNotices` – if set to true, the bot will announce join/part/quit messages (if not set, overrides below)
* `announceSelfJoin` – if set to true, the bot *will* announce itself (by default it skips any join events involving itself)

There are tests to ensure it works appropriately, and manual testing seems to show it working well. I hope this PR has more of a chance of getting merged?

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -325,7 +325,7 @@ class Bot {
    * @param   channel   The channel to send to
    * @param   text      The text to send to discord
    **/
-  sendToDiscord(channel, text) {
+  sendSpecialToDiscord(channel, text) {
       const discordChannelName = this.invertedMapping[channel.toLowerCase()];
       if (discordChannelName) {
       // #channel -> channel before retrieving:
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -256,8 +256,8 @@ class Bot {
     }
   }
 
-  sendToDiscord(author, channel, text) {
-    const discordChannelName = this.invertedMapping[channel.toLowerCase()];
+  getDiscordChannelFor(ircChannel) {
+    const discordChannelName = this.invertedMapping[ircChannel.toLowerCase()];
     if (discordChannelName) {
       // #channel -> channel before retrieving and select only text channels:
       const discordChannel = discordChannelName.startsWith('#') ? this.discord.channels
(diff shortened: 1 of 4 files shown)


Throne3d:
Hmm, coverage decreased somewhere. I'll look into fixing that.

Throne3d:
It looks like the only lines missed now are `if (discordChannelName) {` (a branch, `[[27,1]]`) and `.filter(c => c.type === 'text')`, same as in the regular `sendToDiscord` function.

Throne3d:
The test description of `should not send messages to discord if the channel isn't in the channel mapping` was inaccurate, seeing as how `"#notinchannel": "#otherIRC"` is in fact in the channel mapping. I've updated the test name and created an *actual* test for a channel not in the mapping (`#no-irc`). (This has a side benefit of testing `getDiscordChannelFor` more thoroughly and hence overall increasing coverage.)

lib/bot.js
```diff
@@ -98,6 +100,24 @@ class Bot {
       this.sendToDiscord(author, to, `*${text}*`);
     });
 
+    this.ircClient.on('join', (channel, nick) => {
+      if (this.disableJoinPartMsgs) return;
+      if (nick === this.nickname && !this.announceSelfJoin) return;
```
ekmartin: Didn't know you'd receive join events for yourself. Do we need a configuration option for this though? There's already quite a lot of configuration possible, so maybe it would be better to just keep this off.

lib/bot.js
```diff
@@ -200,33 +220,50 @@ class Bot {
...
-      discordChannel.sendMessage(withAuthor);
-    }
+    // Add bold formatting:
+    const withAuthor = `**<${author}>** ${withMentions}`;
+    logger.debug('Sending message to Discord', withAuthor, channel, '->', `#${discordChannel.name}`);
+    discordChannel.sendMessage(withAuthor);
+  }
+
+  /* Sends a message to Discord exactly as it appears */
+  sendSpecialToDiscord(channel, text) {
```
ekmartin: Hm, is there any reason this can't just use `sendToDiscord` as well?

lib/bot.js
```diff
@@ -32,6 +32,8 @@ class Bot {
     this.commandCharacters = options.commandCharacters || [];
     this.ircNickColor = options.ircNickColor !== false; // default to true
     this.channels = _.values(options.channelMapping);
+    this.disableJoinPartMsgs = options.disableJoinPartMsgs;
```
ekmartin: Thanks for adding a configuration option! I think it would be better to default it to off though, as that's the current behavior. [slack-irc](https://github.com/ekmartin/slack-irc) uses an object for this, so you can choose to enable joins and leaves selectively:
image

I'm not sure if that's really necessary though, a single option might be more than enough. Maybe just `ircStatusNotices: true`.

lib/bot.js
RoyCurtis: Personally, I found this useful to have on. If the bot had to reconnect for some reason (e.g. config change, lost connection to IRC), the self join message acts as a ping to others in the channel, that the IRC bridge is re-established and ready.

That said, it's currently only half-useful as disconnects are not mentioned in-channel, meaning users won't be aware that the bridge is suddenly unavailable for whatever reason. ^1

----

^1 This is not obvious in Discord; users don't think to check that the bot is online in the right sidebar

ekmartin: I see. Could keep it as a non-documented option, but as you're saying I think it might make more sense to print something like `The bot has joined the IRC channel.` then. Not sure how you'd be able to announce if the bridge is down though, as I think most issues with crashes have been on the Discord side (i.e. timing out and not reconnecting), but I guess you could at least print if the IRC connection goes offline. Seems a bit of an edge case though.

RoyCurtis: Indeed; in my edge case, I want to avoid the rare instances where users may start having one-sided conversations into the void. But, so far, that has yet to happen. So maybe it is not a big enough deal :)

ekmartin: In that case, wouldn't it make more sense to save up messages and send them when the IRC connection comes online? Maybe not if it's more than say 10 messages, but in your case I guess it could be a useful feature. Might not be worth the extra code though.

lib/bot.js
Throne3d: I was mostly just doing this since the other PR did – should I add a boolean flag to display it differently, or…?

ekmartin: Never mind, I was a bit confused since I forgot that we append usernames to the messages even when sent to discord (slack-irc doesn't). The current approach makes sense. Could maybe call it something like `sendExactToDiscord` though. Also, `getDiscordChannelFor` -> `findDiscordChannel`/`getDiscordChannel` or something like that?

lib/bot.js
Throne3d: I've converted it to a single `ircStatusNotices` config.

lib/bot.js
Throne3d: Renamed to `sendExactToDiscord` and `findDiscordChannel`.

lib/bot.js
Throne3d: I'm not sure what to do in response to this – should I leave it as is for now, or…?

Throne3d:
(Rebased onto master, rebased to change the config `disableJoinPartMsgs` → (!)`ircStatusNotices` everywhere, committed a change for `sendSpecialToDiscord` → `sendExactToDiscord` and `getDiscordChannelFor` → `findDiscordChannel`.)

ekmartin:
Looks good! Wanna add a line to the advanced example configuration in the README with `ircStatusNotices`?

Throne3d:
Done (and fixed a merge conflict after #204 was merged).

Throne3d:
Partially solves #92?

ekmartin:
Thanks! I'll publish a release with the recent changes in the next few days.

PR #208, opened 2017-03-26
Convert username mentions even if nickname set

Throne3d:
In Discord, using `@username` pings the person with the username even if they currently have their nickname set to something else. As nicknames can have non-ASCII characters in them (or rather, characters that constitute a 'boundary' per the `/@[^\s]+\b/g` regex to match @-mentions of usernames), and usernames don't tend to (or perhaps can't), enabling username mentions even when a nickname is set would allow IRC users to still mention users with such nicknames.

I can't immediately think of a use-case where this distinction would be useful (or expected), and as Discord itself handles it differently I don't think this is the correct behavior.

In addition, I think the test for this behavior (not creating a mention if nickname differs from username) was broken. See where `'Hello, @username!'` becomes `'Hello, @testuser!'` below – it wasn't in fact referencing an existent user, so this was instead testing that mentioning a nonexistent user doesn't change the mention text.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -213,10 +213,7 @@ class Bot {
 
         const user = this.discord.users.find('username', search);
         if (user) {
-          const nickname = guild.members.get(user.id).nickname;
-          if (!nickname || nickname === search) {
-            return user;
-          }
+          return user;
         }
 
         return match;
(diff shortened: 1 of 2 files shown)


Throne3d:
(It doesn't look like I've actually made any lines uncovered by changing these tests, just that some 'tested' lines have been removed.)

ekmartin:
That sounds reasonable. Thank you for a series of high quality pull requests! I've added you as a collaborator to the project. Feel free to help review PRs if you want to and have time.

ekmartin:

PR #214, opened 2017-04-04
Fix IRC quit messages sending to all channels by tracking users

### LINKED ISSUE #213: IRC quit messages are sent to all channels (erroneously)
If an IRC user is in a subset of IRC channels (say, `#a` and `#b`, but the bot is also connected to `#c` and `#d`), when they quit, the quit messages will be sent to all channels that `#a`, `#b`, `#c` and `#d` map to.

This seems to be a problem in Node-IRC (or perhaps it's intended behavior and it's just unclear, since `#c` and `#d` are on the same server). I reproduced it on Freenode and PSIGenix, so it doesn't seem to be limited to one weird IRC server.

The docs say:

> `'quit'`
> `function (nick, reason, channels, message) { }`
> Emitted when a user disconnects from the IRC, leaving the specified array of channels. See the raw event for details on the message object.

For some reason, the list of channels includes all the channels the bot is in, instead of just the channels the user left from. I'm not sure how to resolve this – somehow keep a list of users in each channel?

Throne3d:
Keep a list of users in each channel, using the `names` event in addition to the `join`/`part` events, so as to prevent spamming unnecessary channels when the `quit` event is raised with all channels from a server.

Fixes #213.

I'm not sure if this is a bug with node-irc (the quit event shouldn't be sent for all channels on that server, just the channels the user is in) or if it's intended behavior (the quit event is sent 'for the server', and so for all channels on it). Either way this is a workaround to make the new functionality… less spammy. I might report it there but they don't seem to have very active development.

This also rewrites a test name to make it fit into the syntax ("Bot Events should…") and reflect the new functionality: "should be possible to get the bot to announce itself joining" → "should announce the bot itself when config enabled".

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -57,6 +57,9 @@ class Bot {
     // ircChannel: IRC channel (e.g. #irc)
     this.formatDiscord = this.format.discord || '**<{$author}>** {$withMentions}';
 
+    // Keep track of { channel => [list, of, usernames] } for ircStatusNotices
+    this.channelUsers = {};
+
     this.channelMapping = {};
 
     // Remove channel passwords from the mapping and lowercase IRC channel names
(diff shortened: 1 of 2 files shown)


ekmartin:
I can't think of a better way to fix this either - seems like it's a TODO in the node-irc codebase: https://github.com/martynsmith/node-irc/blob/master/lib/irc.js#L587-L592

I'm not sure if it's strictly worth it, but I think the implementation here is clean enough that we should include it. Maybe use a [set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) instead of an array for the user lists though? That way you can use `.add` and `.delete` instead of needing to find the index etc.

Throne3d:
It doesn't even look like in their code they're trying to match if the user is in that channel – they just delete the user from the channel (whether they were there or not) and proceed to add it to the list, when I think they should be guarding by whether the object's key exists?

I've updated this to use sets – thanks, didn't realize they existed in JavaScript. (Or, well, ECMAScript 6?)

lib/bot.js
```diff
@@ -125,21 +128,36 @@ class Bot {
...
+        return;
+      }
+      this.channelUsers[channel].delete(nick);
       this.sendExactToDiscord(channel, `*${nick}* has left the channel (${reason})`);
     });
 
     this.ircClient.on('quit', (nick, reason, channels) => {
       if (!this.ircStatusNotices || nick === this.nickname) return;
       channels.forEach((channel) => {
+        if (!this.channelUsers[channel].delete(nick)) return;
```
ekmartin: clever use of delete's return value :)

ekmartin:
Ye, it indeed doesn't look like it attempts to find the correct username.
There's a couple of PRs to node-irc for the issue: https://github.com/martynsmith/node-irc/pull/376 and https://github.com/martynsmith/node-irc/pull/359

If they ever get merged we can remove this I guess.

ekmartin:
Released in [2.3.1](https://github.com/reactiflux/discord-irc/blob/master/CHANGELOG.md#231---2017-04-05) - thanks!

PR #218, opened 2017-04-10
Warn if a part/quit is received and no channelUsers is set

### LINKED ISSUE #216: Bot crashes when user leaves on IRC
So after my stupidness in installing the distro's default for node.js (4.2.3) instead of atleast 6, I have found some other issue.

Whenever a user parts the IRC channel, the bot throws an error and crashes with 
`Cannot read property 'delete' of undefined.`

Below is the Text I can see in Console.
https://gist.github.com/Abchrisabc/3ba0f7ee2778127252b215932d3ec0a5

Throne3d:
Attempts to fix #216's crash but is unlikely to fix the underlying issue (that somehow the client isn't saving
a list of nicks in a channel, either because the server isn't sending a names event, or the node-irc library is processing it poorly, or otherwise).

I'm not sure this should be merged until a proper fix is found, but this shouldn't damage anything anyway.

I couldn't produce the actual crash on the particular server so it might be related to the particular channel in use, but this should hopefully work as a stopgap measure so the bot can be used with status messages (well, joins and probably parts) on such channels, while still alerting when weird behavior is going on.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -146,7 +146,11 @@ class Bot {
         delete this.channelUsers[channel];
         return;
       }
-      this.channelUsers[channel].delete(nick);
+      if (this.channelUsers[channel]) {
+        this.channelUsers[channel].delete(nick);
+      } else {
+        logger.warn(`No channelUsers found for ${channel} when ${nick} parted.`);
+      }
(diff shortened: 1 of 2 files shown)


Throne3d:
Rebased onto master, fixed a merge conflict.

PR #219, opened 2017-04-13
Fix ircStatusNotices when channels are not lowercase

### LINKED ISSUE #216: Bot crashes when user leaves on IRC

Throne3d:
Should fix #216, by lowercasing the channel name when it receives each event. This also adds debug messages to make it easier to see when the bot is actually receiving (and presumably processing) these events.

(When I tested it with the #NovaMC channel, it did in fact crash on master but did not on this branch.)

The added tests fail on master – the first, obviously, because master is not trying to lowercase the channel name, and the second reproduces this particular example with fake events, demonstrating the actual failure of the old code.

It may still be useful to merge #218, in case there are other circumstances that could produce this (such as a race condition for the events, which I think is possible though maybe not likely).

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -125,19 +125,24 @@ class Bot {
       this.sendToDiscord(author, to, `*${text}*`);
     });
 
-    this.ircClient.on('join', (channel, nick) => {
+    this.ircClient.on('join', (channelName, nick) => {
+      logger.debug('Received join:', channelName, nick);
       if (!this.ircStatusNotices) return;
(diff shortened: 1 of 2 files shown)


ekmartin:
Good catch - thanks.

Issue #15, opened 2015-12-03
Support for multi-server channels


billimek:
As a client, I want to be able to 'map' channels belonging to multiple different discord servers instead of a single discord server so that I can map across multiple discord servers per account.

billimek:
For example:

```
[
  // Bot 1 (minimal configuration):
  {
    "nickname": "test2",
    "server": "irc.testbot.org",
    "discordEmail": "user@mail.com",
    "discordPassword": "letmein123",
    "channelMapping": {
      "discord_server #general": "#new-irc-channel" // add discord server name before discord server
      "discord_server2 #general": "#other-irc-channel" // add discord server name before discord server
    }
  }
]
```

ekmartin:
With support for this I'm not sure if there's any point in having support for multiple bot users at all. It's there since this is really just a fork of https://github.com/ekmartin/slack-irc. Something like this wouldn't be possible with slack-irc, since it uses Slack's bot users (which don't authenticate like regular users, and are connected to one server). For Discord however I think what you're suggesting might be a better approach. 

One thing that's a bit negative is that it increases the gap between slack-irc and discord-irc. I've been thinking about ways to abstract away most of the modules' code into a general library, because it's kind of a pain to have so much duplicate code to maintain. However, that would definitely make it harder to have differences like what you're suggesting here (in addition to the existing differences), so I'm not sure if it's worth it.

I'm not sure if I'll have time implementing what you're suggesting in the close future though, so it's up for grabs if anyone wants to take a shot at it.

billimek:
Submitted a pull request for this.
https://github.com/reactiflux/discord-irc/pull/18

keenerb:
This is an important feature.  Please look into supporting multiple servers per user account; this is unusable for me right now because I have four servers linked to the same Discord account, and this is unable to connect to three of them.

CallMeM:
I support this, but maybe not with this syntax, maybe a per server configuration alongside the default unique name config? Or maybe a way to alias a each server's channels to a new name? I don't know how possible the last one is, though.

mcmonkey4eva:
Pinging in that I also could use this feature.

demiters:
Yup. Want this also.

Throne3d:
This seems to now be doable, using channel IDs? Where channel names are ambiguous, the bot will just end up using one or the other (which is obviously not the best behavior, but is hardly damaging), but if you use channel IDs (not ambiguous), it seems to work fine with two servers.

[...]

PR #222, opened 2017-04-13
Add documentation for Discord channel IDs and the custom formatting

Throne3d:
I think this is pretty clear from the title – it adds documentation for the new discord channel ID feature (which, as mentioned [here](https://github.com/reactiflux/discord-irc/issues/15#issuecomment-294032226), has not been clearly documented) and also the custom formatting feature.

This may not cover everything or it might be too detailed to fit into that config, in which case I'm plenty happy to move it somewhere more relevant, optionally with more details about the individual intricacies. Feel free to modify the style / phrasing (or request I do so) if you'd rather it be done otherwise – I tried to have it fit into the surrounding text, but may not have done so sufficiently.


ekmartin:
Looks good!

PR #230, opened 2017-05-02
Use webhooks to reflect irc usernames + avatars

### LINKED ISSUE #128: use webhooks to mirror users usernames
using webhooks it's possible for the bot to submit a username with each request, this could reduce the devide between discord and irc.

Fiaxhs:
This fixes #128 

The discord bot gets renamed with the irc nickname, and use the account avatar if there's a matching username in the room.

discord

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -34,10 +34,17 @@ class Bot {
     this.channels = _.values(options.channelMapping);
 
     this.channelMapping = {};
+    this.webhooks = {};
 
     // Remove channel passwords from the mapping and lowercase IRC channel names
     _.forOwn(options.channelMapping, (ircChan, discordChan) => {
       this.channelMapping[discordChan] = ircChan.split(' ')[0].toLowerCase();
+    });
+
+    // Extract id and token from Webhook urls
+    _.forOwn(options.webhooks, (url, chan) => {
+      const [id, token] = url.split('/').slice(-2);
+      this.webhooks[chan] = { id, token };
     });
 
     this.invertedMapping = _.invert(this.channelMapping);
@@ -47,6 +54,9 @@ class Bot {
   connect() {
     logger.debug('Connecting to IRC and Discord');
     this.discord.login(this.discordToken);
+    _.forOwn(this.webhooks, (webhook, channel) => {
+      this.webhooks[channel].client = new discord.WebhookClient(webhook.id, webhook.token);
+    });
 
     const ircOptions = {
       userName: this.nickname,
@@ -151,7 +161,9 @@ class Bot {
   sendToIRC(message) {
     const author = message.author;
     // Ignore messages sent by the bot itself:
-    if (author.id === this.discord.user.id) return;
+    if (author.id === this.discord.user.id ||
+        Object.keys(this.webhooks).some(chan => this.webhooks[chan].id === author.id)
+      ) return;
 
     const channelName = `#${message.channel.name}`;
     const ircChannel = this.channelMapping[channelName];
@@ -189,6 +201,16 @@ class Bot {
     }
   }
 
+  getDiscordAvatar(nick) {
+    const user = this.discord.guilds.first().members.find(
+      member => member.user.username.toLowerCase() === nick.toLowerCase()
+    );
+    if (user) {
+      return user.user.avatarURL;
+    }
+    return null;
+  }
+
   sendToDiscord(author, channel, text) {
     const discordChannelName = this.invertedMapping[channel.toLowerCase()];
     if (discordChannelName) {
@@ -200,6 +222,16 @@ class Bot {
       if (!discordChannel) {
         logger.info('Tried to send a message to a channel the bot isn\'t in: ',
           discordChannelName);
+        return;
+      }
+      if (this.webhooks[discordChannelName]) {
+        this.webhooks[discordChannelName].client.sendMessage(
+          text,
+          {
+            username: author,
+            text,
+            avatarURL: this.getDiscordAvatar(author)
+          }).catch(logger.error);
         return;
       }
 
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -367,7 +367,7 @@ class Bot {
         }
         const nickLowerCase = nick.toLowerCase();
         return member.user.username.toLowerCase() === nickLowerCase
-          || member.nickname.toLowerCase() === nickLowerCase;
+          || (member.nickname && member.nickname.toLowerCase() === nickLowerCase);
       };
 
     // Try to find exact matching case
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -322,6 +322,8 @@ class Bot {
         .find('name', discordChannelName.slice(1)) : this.discord.channels.get(discordChannelName);
 
       if (!discordChannel) {
+        logger.info('Tried to send a message to a channel the bot isn\'t in: ',
+          discordChannelName);
         return null;
       }
       return discordChannel;
@@ -337,13 +339,15 @@ class Bot {
         return webhook;
       }
     }
-    logger.info('Tried to send a message to a channel the bot isn\'t in or without webhook attached: ',
-      discordChannelName);
     return null;
   }
 
   sendToDiscord(author, channel, text) {
     const discordChannel = this.findDiscordChannel(channel);
+    if (!discordChannel) {
+      return;
+    }
+
 
     // Convert text formatting (bold, italics, underscore)
     const withFormat = formatFromIRCToDiscord(text);
@@ -369,6 +373,20 @@ class Bot {
       return match;
     });
 
+    // Webhooks first
+    const webhook = this.findWebhook(channel);
+    if (webhook) {
+      logger.debug('Sending message to Discord via webhook', withMentions, channel, '->', `#${discordChannel.name}`);
+      webhook.client.sendMessage(
+        withMentions,
+        {
+          username: author,
+          text,
+          avatarURL: this.getDiscordAvatar(author)
+        }).catch(logger.error);
+      return;
+    }
+
     const patternMap = {
       author,
       text: withFormat,
@@ -381,19 +399,6 @@ class Bot {
     // Use custom formatting from config / default formatting with bold author
     const withAuthor = Bot.substitutePattern(this.formatDiscord, patternMap);
     logger.debug('Sending message to Discord', withAuthor, channel, '->', `#${discordChannel.name}`);
-    if (!discordChannel) {
-      const webhook = this.findWebhook(channel);
-      if (webhook) {
-        webhook.client.sendMessage(
-          withAuthor,
-          {
-            username: author,
-            text,
-            avatarURL: this.getDiscordAvatar(author)
-          }).catch(logger.error);
-      }
-      return;
-    }
     discordChannel.sendMessage(withAuthor);
   }
 
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -344,10 +344,7 @@ class Bot {
 
   sendToDiscord(author, channel, text) {
     const discordChannel = this.findDiscordChannel(channel);
-    if (!discordChannel) {
-      return;
-    }
-
+    if (!discordChannel) return;
 
     // Convert text formatting (bold, italics, underscore)
     const withFormat = formatFromIRCToDiscord(text);
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -323,11 +323,9 @@ class Bot {
   }
 
   getDiscordAvatar(nick) {
-    const user = this.discord.guilds.first().members.find(
-      member => member.user.username.toLowerCase() === nick.toLowerCase()
-    );
+    const user = this.discord.users.find('username', nick);
     if (user) {
-      return user.user.avatarURL;
+      return user.avatarURL;
     }
     return null;
   }
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -361,13 +361,13 @@ class Bot {
   getDiscordAvatar(nick, channel) {
     const guildMembers = this.findDiscordChannel(channel).guild.members;
     const findByNicknameOrUsername = caseSensitive =>
-      (value) => {
+      (member) => {
         if (caseSensitive) {
-          return value.username === nick || value.nickname === nick;
+          return member.user.username === nick || member.nickname === nick;
         }
         const nickLowerCase = nick.toLowerCase();
-        return value.username.toLowerCase() === nickLowerCase
-          || value.nickname.toLowerCase() === nickLowerCase;
+        return member.user.username.toLowerCase() === nickLowerCase
+          || member.nickname.toLowerCase() === nickLowerCase;
       };
 
     // Try to find exact matching case
@@ -380,7 +380,7 @@ class Bot {
 
     // No matching user or more than one => no avatar
     if (users && users.size === 1) {
-      return users.first().avatarURL;
+      return users.first().user.avatarURL;
     }
     return null;
   }
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -270,7 +270,7 @@ class Bot {
     // Ignore messages sent by the bot itself:
     if (author.id === this.discord.user.id ||
         Object.keys(this.webhooks).some(chan => this.webhooks[chan].id === author.id)
-      ) return;
+    ) return;
 
     const channelName = `#${message.channel.name}`;
     const ircChannel = this.channelMapping[message.channel.id] ||
@@ -329,14 +329,6 @@ class Bot {
     }
   }
 
-  getDiscordAvatar(nick) {
-    const user = this.discord.users.find('username', nick);
-    if (user) {
-      return user.avatarURL;
-    }
-    return null;
-  }
-
   findDiscordChannel(ircChannel) {
     const discordChannelName = this.invertedMapping[ircChannel.toLowerCase()];
     if (discordChannelName) {
@@ -362,6 +354,33 @@ class Bot {
       if (webhook) {
         return webhook;
       }
+    }
+    return null;
+  }
+
+  getDiscordAvatar(nick, channel) {
+    const guildMembers = this.findDiscordChannel(channel).guild.members;
+    const findByNicknameOrUsername = caseSensitive =>
+      (value) => {
+        if (caseSensitive) {
+          return value.username === nick || value.nickname === nick;
+        }
+        const nickLowerCase = nick.toLowerCase();
+        return value.username.toLowerCase() === nickLowerCase
+          || value.nickname.toLowerCase() === nickLowerCase;
+      };
+
+    // Try to find exact matching case
+    let users = guildMembers.filter(findByNicknameOrUsername(false));
+
+    // Now let's search case insensitive.
+    if (!users) {
+      users = guildMembers.filter(findByNicknameOrUsername(true));
+    }
+
+    // No matching user or more than one => no avatar
+    if (users && users.size === 1) {
+      return users.first().avatarURL;
     }
     return null;
   }
@@ -430,7 +449,7 @@ class Bot {
         {
           username: author,
           text,
-          avatarURL: this.getDiscordAvatar(author)
+          avatarURL: this.getDiscordAvatar(author, channel)
         }).catch(logger.error);
       return;
     }
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -36,6 +36,7 @@ class Bot {
     this.channels = _.values(options.channelMapping);
     this.ircStatusNotices = options.ircStatusNotices;
     this.announceSelfJoin = options.announceSelfJoin;
+    this.webhookOptions = options.webhooks;
 
     // "{$keyName}" => "variableValue"
     // author/nickname: nickname of the user who sent the message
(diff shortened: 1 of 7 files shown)


Throne3d:
Could you add tests for `getDiscordAvatar` and `findWebhook`, as well as ensuring that it uses a webhook in `sendToDiscord` when one is available (and then does not go on to send a regular message)? I can probably advise on how to do this if you're unsure.

Fiaxhs:
Sure! I'll do this soon.

Fiaxhs:
Yes, I'm a bit short on free time these days. I'll try to produce something
this week-end :)

Fiaxhs:
Finally got some time to write tests, sorry for the delay :)

Kakkela:
I would like to add this as a warning if this gets pushed to master.
https://trello.com/c/6ZZTEBJw/278-webhook-incorrectly-stacks-message-authors-on-android-client

So basically currently this wont be really readable on android client because of message merging under one author incorrectly.

maorninja:
Aw man, you already made it work with latest commit. Oh well. I'll delete my fork now.

Throne3d:
Testing this locally, it seems to work well, and the tests added should help to ensure it keeps this functionality.

Assuming @ekmartin is okay with it, I think this is good to go.

(I'm slightly worried about IRC users masquerading as authenticated Discord users by changing to their usernames, but I don't think this is readily fixable, and the messages have 'BOT' marked on them, and it presents a problem on the IRC end as is anyway, so this shouldn't actually make it much if at all worse.)

Fiaxhs:
Just added a note on webhooks. However feel free to change it, even the image. I feel quite uncomfortable writing the readme for a project that isn't mine!

Throne3d:
That looks like an accurate and useful summary, so it should be fine! Thanks.

maorninja:
@Fiaxhs Can you make this case-insensitive?

lib/bot.js
```diff
@@ -310,6 +322,14 @@ class Bot {
...
   }
 
+  getDiscordAvatar(nick) {
+    const user = this.discord.users.find('username', nick);
+    if (user) {
+      return user.avatarURL;
+    }
+    return null;
+  }
+
```
Throne3d: I don't think case insensitivity should be a blocker but it does make me wonder what happens if two Discord users have the same display name, as this seems doable, or the same name ignoring case. This is a problem elsewhere already, I guess, such as when an IRC user mentions a Discord user.

This function also currently checks by username as opposed to nickname (display name), which might be unexpected. This should maybe:

1. Check by nickname and username
2. Check by the above, case-insensitively, if no match found
3. If more than one found, default to no fancy icon

That might be more complicated than necessary, but I think it would suffice for most or all use-cases. It should at least check nicknames?

lib/bot.js
```diff
@@ -373,6 +404,20 @@ class Bot {
...
+      webhook.client.sendMessage(
+        withMentions,
+        {
+          username: author,
+          text,
+          avatarURL: this.getDiscordAvatar(author)
+        }).catch(logger.error);
+      return;
+    }
+
```
Throne3d: … I also failed to notice that this seems to discard the custom formatting when sending with a webhook; I think the 'Webhooks first' code should go after the `const withAuthor…` bit, here, and use `withAuthor` instead of `withMentions`.

Throne3d:
(Sorry I didn't notice this before.)

maorninja:
Hey @Fiaxhs, can you update this to the latest commit?

Fiaxhs:
Yes, I'll update my PR.
Also, I'm changing the way I match IRC nicknames and users nicknames, didn't fully understood the concept of guilds at the time.

Throne3d:
The changes look good (the code, that is) but I can't test it on an actual Discord server right now. @ekmartin, opinions?

Fiaxhs:
Oh right, I forgot to test it on my server, was relying on tests solely for
today's commits. I'll check asap (this weekend).

Throne3d:
[Aside: as a note for future, this is going to behave weirdly with #277, probably by sending actions without any formatting if they don't conflict. When they're both merged I think the code should be modified – probably in whichever PR is outstanding last – to wrap text like `_${text}_` if webhooks are in use; that's how Discord handles `/me` actions from Discord users and the webhook message functionality seems designed to make it look as much like a regular user as possible.]

Could you edit the tests to use `addUser` instead of setting the guild members directly like you have? This will fix the weird `<@undefined>` thing by converting it to an ID, as well as (hopefully) ensuring if we modify the code later to retrieve the Discord user elsewhere it will be less likely to break tests. (The reason I use different IDs is so they don't overlap in these tests, in case there's some erroneous thing relating to a particular ID.)

I can't seem to figure out how to push to the PR myself, if that's in fact enabled, which is why I haven't added the commit already. Here's the patch for how I'd do it, which seems to pass tests:

```patch
diff --git a/test/bot.test.js b/test/bot.test.js
index 9da4d54..b607a2f 100644
--- a/test/bot.test.js
+++ b/test/bot.test.js
@@ -836,37 +836,42 @@ describe('Bot', function () {
     const newConfig = { ...config, webhooks: { '#discord': 'https://discordapp.com/api/webhooks/id/token' } };
     const bot = new Bot(newConfig);
     bot.connect();
-    const user = new discord.User(this.bot.discord, { username: 'Nick', avatar: 'avatarURL' });
-    this.guild.members.set(1, { user, nickname: 'Different' });
-    this.bot.getDiscordAvatar('Nick', '#irc').should.equal('/avatars/<@undefined>/avatarURL.png?size=2048');
+    const userObj = { id: 123, username: 'Nick', avatar: 'avatarURL' };
+    const memberObj = { nickname: 'Different' };
+    this.addUser(userObj, memberObj);
+    this.bot.getDiscordAvatar('Nick', '#irc').should.equal('/avatars/123/avatarURL.png?size=2048');
   });

   it('should find a matching username, case insensitive, when looking for an avatar', function () {
     const newConfig = { ...config, webhooks: { '#discord': 'https://discordapp.com/api/webhooks/id/token' } };
     const bot = new Bot(newConfig);
     bot.connect();
-    const user = new discord.User(this.bot.discord, { username: 'nick', avatar: 'avatarURL' });
-    this.guild.members.set(1, { user, nickname: 'Different' });
-    this.bot.getDiscordAvatar('Nick', '#irc').should.equal('/avatars/<@undefined>/avatarURL.png?size=2048');
+    const userObj = { id: 124, username: 'nick', avatar: 'avatarURL' };
+    const memberObj = { nickname: 'Different' };
+    this.addUser(userObj, memberObj);
+    this.bot.getDiscordAvatar('Nick', '#irc').should.equal('/avatars/124/avatarURL.png?size=2048');
   });

   it('should find a matching nickname, case sensitive, when looking for an avatar', function () {
     const newConfig = { ...config, webhooks: { '#discord': 'https://discordapp.com/api/webhooks/id/token' } };
     const bot = new Bot(newConfig);
     bot.connect();
-    const user = new discord.User(this.bot.discord, { username: 'Nick', avatar: 'avatarURL' });
-    this.guild.members.set(1, { user, nickname: 'Different' });
-    this.bot.getDiscordAvatar('Different', '#irc').should.equal('/avatars/<@undefined>/avatarURL.png?size=2048');
+    const userObj = { id: 125, username: 'Nick', avatar: 'avatarURL' };
+    const memberObj = { nickname: 'Different' };
+    this.addUser(userObj, memberObj);
+    this.bot.getDiscordAvatar('Different', '#irc').should.equal('/avatars/125/avatarURL.png?size=2048');
   });

   it('should not return an avatar with two matching usernames when looking for an avatar', function () {
     const newConfig = { ...config, webhooks: { '#discord': 'https://discordapp.com/api/webhooks/id/token' } };
     const bot = new Bot(newConfig);
     bot.connect();
-    const user1 = new discord.User(this.bot.discord, { username: 'common', avatar: 'avatarURL' });
-    const user2 = new discord.User(this.bot.discord, { username: 'Nick', avatar: 'avatarURL' });
-    this.guild.members.set(1, { user: user1, nickname: 'Different' });
-    this.guild.members.set(2, { user: user2, nickname: 'common' });
+    const userObj1 = { id: 126, username: 'common', avatar: 'avatarURL' };
+    const userObj2 = { id: 127, username: 'Nick', avatar: 'avatarURL' };
+    const memberObj1 = { nickname: 'Different' };
+    const memberObj2 = { nickname: 'common' };
+    this.addUser(userObj1, memberObj1);
+    this.addUser(userObj2, memberObj2);
     chai.should().equal(this.bot.getDiscordAvatar('common', '#irc'), null);
   });
 });
```

As well as this, could you make sure that it doesn't crash when a user doesn't have a nickname (I just found it to do that when I tested it and there was no match for an IRC user):

```patch
diff --git a/lib/bot.js b/lib/bot.js
index 1388665..233d9a7 100644
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -367,7 +367,7 @@ class Bot {
         }
         const nickLowerCase = nick.toLowerCase();
         return member.user.username.toLowerCase() === nickLowerCase
-          || member.nickname.toLowerCase() === nickLowerCase;
+          || (member.nickname && member.nickname.toLowerCase() === nickLowerCase);
       };

     // Try to find exact matching case
diff --git a/test/bot.test.js b/test/bot.test.js
index b607a2f..f7c5351 100644
--- a/test/bot.test.js
+++ b/test/bot.test.js
@@ -874,4 +874,17 @@ describe('Bot', function () {
     this.addUser(userObj2, memberObj2);
     chai.should().equal(this.bot.getDiscordAvatar('common', '#irc'), null);
   });
+
+  it('should not return an avatar when no users match and should handle lack of nickname, when looking for an avatar', function () {
+    const newConfig = { ...config, webhooks: { '#discord': 'https://discordapp.com/api/webhooks/id/token' } };
+    const bot = new Bot(newConfig);
+    bot.connect();
+    const userObj1 = { id: 128, username: 'common', avatar: 'avatarURL' };
+    const userObj2 = { id: 129, username: 'Nick', avatar: 'avatarURL' };
+    const memberObj1 = {};
+    const memberObj2 = { nickname: 'common' };
+    this.addUser(userObj1, memberObj1);
+    this.addUser(userObj2, memberObj2);
+    chai.should().equal(this.bot.getDiscordAvatar('nonexistent', '#irc'), null);
+  });
 });
```

After applying these two (well, mainly the latter), it seems to work quite nicely. Thanks for your contribution!

Throne3d:
This looks good to go now. @ekmartin?

Fiaxhs:
Nice! I didn't see the `addUser` function, this would have save me some time.
> I keep finding things to be picky about

Fine by me :)
> Thanks for your contribution!

You're welcome! I use it every day, for once I could bring some value to a project I use.

lib/bot.js
```diff
@@ -65,19 +65,29 @@ class Bot {
...
     this.channelMapping = {};
+    this.webhooks = {};
 
     // Remove channel passwords from the mapping and lowercase IRC channel names
     _.forOwn(options.channelMapping, (ircChan, discordChan) => {
       this.channelMapping[discordChan] = ircChan.split(' ')[0].toLowerCase();
     });
 
+    // Extract id and token from Webhook urls
+    _.forOwn(options.webhooks, (url, chan) => {
```
ekmartin: `chan` -> `channel`

ekmartin: (consistent with your block on line 80)

ekmartin: on second thought, might as well merge this down into the `connect` block and only have one `forOwn(options.webhooks, ...)`?

lib/bot.js
```diff
@@ -258,7 +268,9 @@ class Bot {
   sendToIRC(message) {
     const author = message.author;
     // Ignore messages sent by the bot itself:
-    if (author.id === this.discord.user.id) return;
+    if (author.id === this.discord.user.id ||
+        Object.keys(this.webhooks).some(chan => this.webhooks[chan].id === author.id)
```
ekmartin: channel

lib/bot.js
```diff
@@ -335,6 +347,44 @@ class Bot {
     return null;
   }
 
+  findWebhook(ircChannel) {
+    const discordChannelName = this.invertedMapping[ircChannel.toLowerCase()];
+    if (discordChannelName) {
```
ekmartin: Might as well do: `return discordChannelName && this.webhooks[discordChannelName]`.

lib/bot.js
```diff
@@ -335,6 +347,44 @@ class Bot {
...
+      }
+    }
+    return null;
+  }
+
+  getDiscordAvatar(nick, channel) {
+    const guildMembers = this.findDiscordChannel(channel).guild.members;
+    const findByNicknameOrUsername = caseSensitive =>
+      (member) => {
+        if (caseSensitive) {
```
ekmartin: Are discord nicknames case sensitive or insensitive? Can there be two users with the same username/nickname with different casing? I think we can simplify this logic as long as we know what assumptions we're making.

lib/bot.js
```diff
@@ -390,6 +440,20 @@ class Bot {
...
+    const webhook = this.findWebhook(channel);
+    if (webhook) {
+      logger.debug('Sending message to Discord via webhook', withMentions, channel, '->', `#${discordChannel.name}`);
+      webhook.client.sendMessage(
+        withMentions,
+        {
+          username: author,
+          text,
+          avatarURL: this.getDiscordAvatar(author, channel)
+        }).catch(logger.error);
```
ekmartin: Would the error here not bubble up to our `on('error'` handler? (I don't know)

lib/bot.js
```diff
@@ -390,6 +440,20 @@ class Bot {
       return match;
     });
 
+    // Webhooks first
+    const webhook = this.findWebhook(channel);
+    if (webhook) {
+      logger.debug('Sending message to Discord via webhook', withMentions, channel, '->', `#${discordChannel.name}`);
+      webhook.client.sendMessage(
```
ekmartin: formatting nitpick:
```js
webhook.client.sendMessage(withMentions, {
  text,
  username: author,
  avatarUrl: this.getDiscordAvatar(author, channel)
})
```

(could even do `const avatarURL = this.getDiscordAvatar(author, channel);` first - always optimize for readability)

test/bot.test.js
```diff
@@ -807,4 +810,81 @@ describe('Bot', function () {
     this.sendStub.should.have.been.calledOnce;
     this.sendStub.getCall(0).args.should.deep.equal([msg]);
   });
+
+  it('should create webhooks clients for each webhook url in the config', function () {
```
ekmartin: Lots of tests - nice!

lib/bot.js
Throne3d: Insensitive and yes, which is why I suggested (I think I suggested?) matching case sensitively first and then falling back to case insensitive if there's no match found. (Since IRC users have no way to, in a dropdown, select the correct one of the various case sensitivities – Discord users have this ability since Discord translates the mention in its UI – I thought it best to allow them to force the right one by picking the right case, where relevant.)

Fiaxhs: You can even have two users with the exact same nickname: http://i.imgur.com/ACuYlcS.png + http://i.imgur.com/oD3JEfK.png
(And @maorninja suggested the change :) )

lib/bot.js
Fiaxhs: Webhook clients aren't attached to the main discord client: 

```this.discord = new discord.Client({ autoReconnect: true });```
vs. 
```this.webhooks[channel] = { [...], client: new discord.WebhookClient(id, token);}```
So there's no way errors will bubble up to the main ```on('error'```

I could probably define a function and use it for both error handling, though.

File: lib/bot.js (hunk 6 above)
Comment (Throne3d) [inline lib/bot.js]: (Yes but I suggested [here](https://github.com/reactiflux/discord-irc/pull/230#discussion_r125168767) the logic for how to implement it. :P)

Comment (Fiaxhs) [inline lib/bot.js]: ( right :D )

Comment (maorninja) [inline lib/bot.js]: Personally, I would blame Discord because they allow 2 users with the exact same nickname, but that's just my opinion. I wish there was a bot that could block that.

File: lib/bot.js (hunk 1 above)
Comment (qaisjp) [inline lib/bot.js]: What's the purpose of checking case-sensitively at all? Performance?

**Comment by qaisjp** [comment]
What if we fall back to adorable.io avatars? 

See https://github.com/ekmartin/slack-irc/ and https://github.com/qaisjp/go-discord-irc/commit/859348452f1bb5a47f6189064b5c79ef782e9776

**Comment by qaisjp** [comment]
Also, please send feedback to Discord asking them to put some focus on this! https://trello.com/c/6ZZTEBJw

@Kakkela's https://github.com/reactiflux/discord-irc/pull/230#issuecomment-304867667 shows a major block

File: lib/bot.js (hunk 1 above)
Comment (Throne3d) [inline lib/bot.js]: So that if you have two users with the same nickname barring case, you can pick a particular one to use. (It's also possible to have two users with the exact same nickname, case-sensitively, but then the ambiguity is basically required as opposed to mostly quite avoidable.)

Maybe it's over-much, though – if two users have the same nickname, case-sensitive or not, it's not particularly unreasonable to expect weird behavior around the ambiguity.

**Comment by Throne3d** [comment]
@qiasjp

I'm not sure there's a need to fallback to specific avatars – if the users haven't got Discord accounts with avatars, then they haven't chosen an avatar, so it seems sensible to stick to what Discord does by default? (That is, the standard game controller picture.)

I agree that there should be a warning about the bug when this gets merged. That's a pretty big issue on Discord's end. I don't think it should be a block to us merging this, but instead added as a clear warning in the relevant config area, so users of the individual servers can choose whether to enable it despite this in the meantime (or not).

**Comment by qaisjp** [comment]
@Throne3d yes but the Discord default avatars have a bit of variety. the webhook currently has the same default avatar across usernames.

fyi the only reason I suggested it is because @ekmartin's slack bridge supports it

**Comment by qaisjp** [comment]
Also, something to mention to people reading this issue:

A nice workaround for the Android/web bug is to alternate the webhook per user 

(This wouldn't be suitable for reactiflux/discord-irc, but my purpose build discord-irc bridge bot will do this)

File: lib/bot.js [hunk 10]
```diff
@@ -390,6 +436,19 @@ class Bot {
       return match;
     });
 
+    // Webhooks first
+    const webhook = this.findWebhook(channel);
+    if (webhook) {
+      logger.debug('Sending message to Discord via webhook', withMentions, channel, '->', `#${discordChannel.name}`);
+      const avatarURL = this.getDiscordAvatar(author, channel);
```
qaisjp: During netsplits `author` is nil. This would usually send some message like `<${username}> ${message}` (or something similar, I can't recall. but the message would contain `${}` sent verbatim). Now `this.getDiscordAvatar` crashes with `TypeError: Cannot read property 'toLowerCase' of undefined`

maorninja:
Hey, I have a recommendation for this.

Basically, make the bot have a little command where if you type !ban {user} on discord, you can ban them on IRC.

And how this works? Well, it basically detects the rank that the user who sent it has and checks if he has the "Ban users" perm. If he has it, and the bot on IRC has the perm to do so as well, it'll ban the user off IRC from Discord.

Fiaxhs:
@Kovaelin Here's what my config looks like:

```
{
  "channelMapping": {
    "#machinzone": "#machinzonefr",
    "#crypto": "#machincrypto",
    "#dev": "#machindev"
  }
  [...]
  "webhooks": {
    "#crypto": "https://discordapp.com/api/webhooks/xxxx/XXXXX",
    "#machinzone": "https://discordapp.com/api/webhooks/xxxx/XXXXX",
    "#dev": "https://discordapp.com/api/webhooks/xxxx/XXXXX"
  }
}
```

Kovaelin:
I still can't seem to get the webhook to work. Do I have to change any of the other files? Does it work if I use the Discord channel ID rather than the channel name?

My config file looks like this:
[
  {
    "nickname": "ROBOBOT",
    "server": "IRC.SERVER.BOT",
    "discordToken": "TOkeN",
    "commandCharacters": ["!"],
    "ircStatusNotices": true,
    "ircOptions": {
      "floodProtection": false,
      "encoding": "utf-8"
    },
    "channelMapping": {
      "DISCORDIDNUMBERSA": "#IRC"
    },
    "webhooks": {
      "NUMBERSA": "https://discordapp.com/api/webhooks/COPYPASTEA/COPYPASTEB"
    }
  },

  {
    "nickname": "ROBOBOT2",
    "server": "IRC.SERVER.BOT",
    "discordToken": "TOkeN",
    "channelMapping": {
      "DISCORDIDNUMBERSB": "#IRC2"
    },
    "commandCharacters": ["!"]
  }
]

Fiaxhs:
@Kovaelin Could you open an issue on https://github.com/Fiaxhs/discord-irc instead, please?

ryantheleach:
Still looking for this even with the bug. This would be a huge improvement to quality of life

Kakkela:
Webhook merging bug on Android is now moved to "Claimed fixed, needs verification"
I'm going to test on my bridge today or tomorrow. Not sure if it's fixed on just Alpha channel on Android or what.

https://trello.com/c/6ZZTEBJw/278-webhook-incorrectly-stacks-message-authors-on-android-client

qaisjp:
Omg excited. Please keep us updated :)

Kakkela:
Yes this does now indeed work on Android (all channels) as intended and no unnecessary merging happens.

qaisjp:
And it's fixed on web too!!

ekmartin:
Published in version [2.6.0](https://github.com/reactiflux/discord-irc/blob/master/CHANGELOG.md#260---2018-03-22) of discord-irc - thank you for your patience!

I wonder if the usernames on the Discord side should be suffixed with `(IRC)` to make what's happening clearer though? If you think that's a good idea we can publish a new version later.

qaisjp:
Side note: discord webhooks have a minimum name length of 2 characters, and
having a suffix of `(IRC)` would be an easy countermeasure.

ekmartin:
Opened an issue for it: #377 - PRs welcome!

Kovaelin:
Thank you for this update!

PR #235, opened 2017-05-09
send nick change from irc to discord

Junky2k:
anoy discord user with the nick changes of the irc users

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -128,7 +128,13 @@ class Bot {
     this.ircClient.on('nick', (oldnick, newnick, channels) => {
       channels.forEach((channelName) => {
         const channel = channelName.toLowerCase();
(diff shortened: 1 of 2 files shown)


lib/bot.js
```diff
@@ -125,6 +125,20 @@ class Bot {
       this.sendToDiscord(author, to, `*${text}*`);
     });
 
+    this.ircClient.on('nick', (oldnick, newnick, channels) => {
```
ekmartin: `oldNick` and `newNick` maybe? Nitpick :)

Throne3d:
The only thing I would be wary about is if the library sends the list of channels containing *every* channel, like it does for quit, in which case this (I think) will send to all channels, even ones the user isn't in.

I haven't actually tried it out in practice yet so this might be a nonissue.

Throne3d:
This should work fine, I think.

ekmartin:
Thanks!

PR #241, opened 2017-05-16
check if an user is present in the channel for nickchange

Junky2k:
i only check if the channel has been tracked, not if the user was present in the channel.

i fu**ed up, i'm sorry :D

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -129,9 +129,11 @@ class Bot {
       channels.forEach((channelName) => {
         const channel = channelName.toLowerCase();
         if (this.channelUsers[channel]) {
-          this.channelUsers[channel].delete(oldNick);
-          this.channelUsers[channel].add(newNick);
-          this.sendExactToDiscord(channel, `*${oldNick}* is now known as ${newNick}`);
+          if (this.channelUsers[channel].has(oldNick)) {
+            this.channelUsers[channel].delete(oldNick);
+            this.channelUsers[channel].add(newNick);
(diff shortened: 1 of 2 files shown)


lib/bot.js
```diff
@@ -129,9 +129,11 @@ class Bot {
...
-          this.channelUsers[channel].delete(oldNick);
-          this.channelUsers[channel].add(newNick);
-          this.sendExactToDiscord(channel, `*${oldNick}* is now known as ${newNick}`);
+          if (this.channelUsers[channel].has(oldNick)) {
+            this.channelUsers[channel].delete(oldNick);
+            this.channelUsers[channel].add(newNick);
+            this.sendExactToDiscord(channel, `*${oldNick}* is now known as ${newNick}`);
+          }
         } else {
           logger.warn(`No channelUsers found for ${channel} when ${oldNick} changed.`);
```
ekmartin: To be honest I don't think we really need the logger warning here, so maybe just fold the two `if`s into one and do `if (this.channelUsers[channel] && this.channelUsers[channel].has(oldNick)) {`?

ekmartin: Or what do you think @Throne3d?

ekmartin:
no worries, thanks for the fix :)

lib/bot.js
Throne3d: The logger warning probably isn't necessary – it was mostly to get around the issue with lowercasing channel names, in the other events.

ekmartin: Eh, I guess not. Just a bit cleaner code, but it'll do.

Issue #197, opened 2017-03-12
Avoid self-message highlighting on IRC


ekmartin:
#193 originally fixed this, but caused issues with some clients (#196). Open for other suggestions.

Sanqui:
Sadly, I don't think there's a catch-all solution.  With a plethora of systems and IRC clients, you're always going to run into issues with some of them.  I would go with @Kroisse's #194 implementation which lets maintainers turn the feature on or off depending on if it works for their particular users.

Sanqui:
Putting the zero width character between the first and second letter only, rather than each of them, like #194 did, should help reduce the collateral damage when the character is visible, at slight risk of triggering more highlights (e.g. for people who have a prefixed `_name`).

Issue #196, opened 2017-03-12
Messages to IRC contain &zwj; in nicknames; Causes issues with some clients (and platforms)


duckspike:
As the title states, when the bot sends messages from Discord to IRC, the formatted username of the Discord chatter contains &zwj; between each letter of the nickname. An example is pictured below.

![screen shot 2017-03-10 at 9 50 56 pm](https://cloud.githubusercontent.com/assets/15951552/23836915/8c1eb3bc-075e-11e7-9400-d149bc90cf8f.png)

It seems to cause issues with push notifications on Google Chrome on Mac OS as well IRC applications on iOS devices. On iOS, the character does not render properly and causes the application to slow down for a length of about 0.3 seconds. A below screenshot proves the issue with the Mutter IRC client for iOS. 

![img_1121](https://cloud.githubusercontent.com/assets/15951552/23836975/6878fcfa-075f-11e7-9d25-a2928cc31368.PNG)

Using the latest version of discord-irc as of 3/10/17 and nodejs 6.10.0 compiled from source.

ekmartin:
That's unfortunate. I've reverted the changes introduced in #193 and 2.2.0, and published a new version (2.2.1). @Sanqui do you have any suggestions for how we might fix this?

PR #243, opened 2017-05-20
Add `ircPreventMention` option

qaisjp:
I have the same username on both Discord and IRC. If I write a message in Discord, I receive a nickalert (a "notification", or a "mention") in IRC, which can be very annoying. This PR adds an option (defaults to "on") to prevent that from happening.

By inserting any zero width character (in this pull request, I use a zero width space) somewhere in the printed username you can prevent a nickalert from occurring, with the name looking exactly the same.

For example, with username `john`, I slice `j`, add a zero-width-space, and add `ohn` to the end. This looks like `john`, but is actually `j?ohn` where `?` is the zero width character. 

For usernames that are one character long, the zero width character is at the end of the username (the code does not crash). Depending on IRC client settings, this may or may not notify the client.

The final commit is required so that `ircNickColor` does not overwrite the `displayUsername` change earlier on by `ircPreventMention`. Colours will remain the same as the code that generates the colour still uses `nickname`.

I've actually tested functionality, and only authored it in the GitHub interface because it's easier than using Vim on my server!

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ First you need to create a Discord bot user, which you can do by following the i
       // {$ircChannel} (e.g. #irc)
     },
     "ircNickColor": false, // Gives usernames a color in IRC for better readability (on by default)
+    "ircPreventMention": false, // Prevents users of both IRC and Discord from being mentioned in IRC when they speak in Discord (on by default)
     // Makes the bot hide the username prefix for messages that start
     // with one of these characters (commands):
     "commandCharacters": ["!", "."],
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -280,7 +280,7 @@ class Bot {
       
       if (this.ircNickColor) {
         const colorIndex = (nickname.charCodeAt(0) + nickname.length) % NICK_COLORS.length;
-        displayUsername = irc.colors.wrap(NICK_COLORS[colorIndex], nickname);
+        displayUsername = irc.colors.wrap(NICK_COLORS[colorIndex], displayUsername);
       }
 
       const patternMap = {
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -275,9 +275,9 @@ class Bot {
       if (this.ircPreventMention) {
         // Prevent users of both IRC and Discord from
         // being mentioned in IRC when they talk in Discord.
-        displayUsername = displayUsername.slice(0, 1) + "\u200B" + displayUsername.slice(1);
+        displayUsername = `${displayUsername.slice(0, 1)}\u200B${displayUsername.slice(1)}`;
       }
-      
+
       if (this.ircNickColor) {
         const colorIndex = (nickname.charCodeAt(0) + nickname.length) % NICK_COLORS.length;
         displayUsername = irc.colors.wrap(NICK_COLORS[colorIndex], displayUsername);
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -33,7 +33,7 @@ class Bot {
     this.discordToken = options.discordToken;
     this.commandCharacters = options.commandCharacters || [];
     this.ircNickColor = options.ircNickColor !== false; // default to true
-    this.ircPreventMention = options.ircPreventMention !== false; // default to true
+    this.ircPreventMention = options.ircPreventMention === true; // default: false
     this.channels = _.values(options.channelMapping);
     this.ircStatusNotices = options.ircStatusNotices;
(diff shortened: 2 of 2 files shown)


qaisjp:
This breaks a bunch of tests... I'll write new tests.

qaisjp:
Should I fix the broken tests by setting the ircPreventMention setting to false in the test config, and then writing one test that checks the on/off option?

ekmartin:
#193 implemented this, but was reverted because of #196. Even though this puts it behind a config option, I don't think it's a good idea to add something that potentially causes issues for IRC clients nonetheless.

Maybe your IRC client supports disabling mentions for a specific channel?

qaisjp:
Ah, that's a shame. I did not notice this had already been implemented.

Also, it seems like one of the issues in #196 (macOS, Chrome, IRCCloud) no longer occur. See below:

image


I use `\u200B` whereas the previous PR used `\u200d`. If I get confirmation on iOS that things are OK, will there be any possibility of this PR being reopened?

I know XChat hates symbols, so I'll do a check there and if XChat doesn't complain, it's unlikely it causes issues on other desktop clients.

Throne3d:
This PR looks like it basically implements [this suggestion](https://github.com/reactiflux/discord-irc/issues/197#issuecomment-286859889) from issue #197, which I think is a reasonable compromise.

I don't know what systems had trouble with the notifications, though – perhaps we should consider that this character could interact differently, and so push it and revert if there are further bug reports?

ekmartin:
I'm not completely sure - it's still sort of a hack. What are some use cases where either disabling highlights for that IRC channel or just leaving the channel altogether isn't enough?

qaisjp:
@ekmartin When you want to use both clients. My Discord server doesn't have all the channels on the IRC server I frequent, so I need to use both IRC and Discord. Having to flip between both clients isn't fun, but replying to whichever I have open at the time is more convenient.

I've been using this edit since the creation of this PR (in a 52 person IRC channel), and today I've just added this bot (with this mod) to another server (in a 76 person IRC channel).

No complaints so far. I'll come back later and give feedback from the bigger server.

qaisjp:
No issues have been reported from the 2nd Discord guild. I'm running it on a third guild now, lets see what they say...

qaisjp:
It's running on three guilds — no issues have been reported.

webvictim:
Would love to see this merged!

Throne3d:
My comment would be – it might be easier to have it merged if it defaults to false? Less objectionable. I don't see a major reason *not* to merge it (I would personally find it useful, and would offer to look after this part of code if necessary).

@ekmartin do you have further input here? I think the major blocker was not seeing an obvious use for it, but I do know some people who prefer (for whatever reason) to be in both Discord and IRC. (Perhaps they like the textual logging IRC can provide, and haven't sorted something out for Discord?) It seems slightly out of scope, but not dramatically, as this sort of thing was merged before, then found to cause problems.

qaisjp:
Obviously I'll have to write some tests here for ircPreventMention, but shouldn't these tests pass if ircPreventMention defaults to false?

melt7777:
What a great feature! How can I use this? Is it an option in the config file? Do you have an ASCII-code (Alt-code) for the blank space? Thanks!

qaisjp:
@melt7777 Yes, just add `"ircPreventMention": true` to the config.

The code is `\u200B`

Throne3d:
@qaisjp You're currently inverting the condition – `options.ircPreventMention !== true` is the same as `!(options.ircPreventMention === true)` is the same as `!options.ircPreventMention`.

You probably want just `this.ircPreventMention = options.ircPreventMention`? (That way it defaults to undefined, and must be explicitly set to `true` to attain a truthy value.)

melt7777:
I pulled the latest from git and then added that to my config, pm2 restart discord-irc, and then it joins, but behavior is the same. Also, when I try Change Nickname on the server and change it from melt to m\u200Belt it just makes it that and does not parse the \u control. Does it have to be my username?

webvictim:
@melt7777 I think you're misunderstanding what the feature does. When set to true, the bot will insert a #\u200B# into every <nick> it sends to IRC to try and avoid highlighting the person. You don't need to change your nick on either IRC or Discord at all.

anirbanmu:
Is there any plan to merge this?

Throne3d:
I'm furthering this in #502 (added a test and also fixed the README), plan is to merge it as soon as Travis passes! Gonna close this, accordingly.

Thanks for the contribution. :)

Issue #250, opened 2017-06-12
Server-specific emoji are not translated into the appropriate codes IRC → Discord


Throne3d:
As in the title, if the server has (say) a custom emoji named `:poof:`, writing this on Discord will produce the relevant picture and be translated on the IRC side to the text.

If someone on the IRC side writes `:poof:`, it will appear as the plaintext instead of being translated to the emoji. This is probably fixable in a similar way to how user, channel and role mentions are currently being handled. (I plan on looking at this in about two weeks if nobody else gets to it first.)

Throne3d:
Fixed in #256.

PR #256, opened 2017-06-29
Translate custom emoji references from IRC to Discord

Throne3d:
When an IRC user uses a custom emoji reference (e.g. :testemoji:), it should be translated into the actual
emoji for Discord users, instead of appearing in plain text.

This also adds tests to ensure other emoji-like phrases do not get converted or cause the bot to crash.

(Discord users enter something such as `:testemoji:`, which gets converted into e.g. `<:testemoji:987654321>` – we already translate this to IRC, stripping the angle brackets and extra digits, but don't translate the other way as I would expect.)

Addresses issue #250.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -371,6 +371,14 @@ class Bot {
       }
 
       return match;
+    }).replace(/:(\w+):/g, (match, ident) => {
+      const guild = discordChannel.guild;
+      const emoji = guild.emojis.find(x => x.name === ident && x.requiresColons);
+      if (emoji) {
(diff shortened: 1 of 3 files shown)


Throne3d:
As a note: I think all server-specific emoji have `requiresColons` true, but in the case that one doesn't, it presumably shouldn't map `:emojiname;`, since the emoji doesn't "require colons". I'm not sure what it would look like, as all the server-specific emoji I can find or create seem to need colons, but I thought it best to require that anyway.

ekmartin:
Cool, thanks!

PR #257, opened 2017-07-01
Use ircClient.nick instead of nickname when guarding against events for current nick

Throne3d:
As the irc library can change the nick of the bot when connecting (e.g. because the nick is currently taken, by some other user), `this.nickname` may not correspond to the current nickname and so when receiving a join event upon joining a channel (prior to the names event), the current guard against this can pass despite it being an event for the bot.

This modifies it to check against ircClient.nick where relevant, and also modifies the test stubs to allow adding the nickname to the config and exposing this functionality in tests.

To see the failure before this patch, enable `ircStatusNotices` and run two copies of the bot; the second should crash upon joining a channel, as it will have a modified nickname (e.g. 'testbot1').

The use of `...args` in the constructor is because it seems writing `constructor(_, nickname, _)` clashes due to same-named parameters, and `constructor(_server, nickname, _ircOptions)` seems to complain about unused variables (though the `_`s are intended to show that this has been acknowledged).

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -144,11 +144,11 @@ class Bot {
     this.ircClient.on('join', (channelName, nick) => {
       logger.debug('Received join:', channelName, nick);
       if (!this.ircStatusNotices) return;
-      if (nick === this.nickname && !this.announceSelfJoin) return;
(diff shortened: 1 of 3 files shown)


ekmartin:
Good catch, thanks!

PR #260, opened 2017-07-01
Skip sending falsy command preludes

### LINKED ISSUE #239: Commands to IRC from Discord without announcement
Ok, so as its currently set by default, command characters (such as !) that are set in the config are prefixed by a line saying Command sent from Discord by name:
I know it the current version, it can be changed to something else, but is there a way to omit that entirely, just have it send the command through?

Throne3d:
Fixes #239 by allowing falsy command preludes to override default, and then manually skipping them in the sendToIrc and sendToDiscord methods.

(The old `||` method for defaulting meant falsy preludes were overridden. It functions properly with just that area fixed, but the other modifications are to prevent sending empty messages, which causes `discord.js` to complain and is probably a bad idea to accidentally do anyway.)

This seems possibly abusable by users to hide who's sending commands from the IRC users' perspective, but it'd have to be done intentionally by the person running the bot, so I think it's acceptable.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -51,7 +51,11 @@ class Bot {
     this.formatURLAttachment = this.format.urlAttachment || '<{$displayUsername}> {$attachmentURL}';
     // "{$keyName}" => "variableValue"
     // side: "Discord" or "IRC"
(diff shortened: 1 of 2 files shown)


ekmartin:
👍

PR #272, opened 2017-07-17
Modify tests to use actual discord Collections and a single guild for clarity

Throne3d:
Instead of stubbing the find methods for the various objects, which has for me become more unwieldy when trying to use other collection methods such as filter, this has the tests use actual collection objects from `discord.js`. This also adds methods `addUser`, `addRole` and `addEmoji` to add these items to the current guild instead of the previous stubbing.

There's been a slight modification in expected result in the tests – for one, when mentioning a user who has a nickname set, the reference is now of the form `<@!###>` instead of `<@###>` (note the `!`). This was changed in https://github.com/hydrabolt/discord.js/commit/51f30d6e4cbe2c4a5d40c3aae471150d92d7ea0a, but since we hadn't actually used the built-in classes we didn't notice the change. (It wouldn't have broken anything, just tests.)

This is mostly done as groundwork for another PR I'm about to submit, but it should help to demystify some of the aspects of the testing here, since that was a little confusing for me when I looked at it first. (There was the guild manually specified on the channel in the helper, and the `createGuildStub()` thing, and I wasn't sure what to modify or where. That sort of thing.)

### DIFF
--- a/test/bot.test.js
+++ b/test/bot.test.js
@@ -20,24 +20,50 @@ describe('Bot', function () {
     useFakeServer: false
   });
 
+  const createGuildStub = () => ({
+    roles: new discord.Collection(),
+    members: new discord.Collection(),
+    emojis: new discord.Collection()
+  });
+
   beforeEach(function () {
     this.infoSpy = sandbox.stub(logger, 'info');
     this.debugSpy = sandbox.stub(logger, 'debug');
     this.errorSpy = sandbox.stub(logger, 'error');
     this.sendStub = sandbox.stub();
-    this.findUserStub = sandbox.stub();
-    this.findRoleStub = sandbox.stub();
-    this.findEmojiStub = sandbox.stub();
+
+    this.discordUsers = new discord.Collection();
     irc.Client = ClientStub;
-    discord.Client = createDiscordStub(
-      this.sendStub, this.findUserStub, this.findRoleStub, this.findEmojiStub
-    );
+    this.guild = createGuildStub();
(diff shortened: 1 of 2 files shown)


test/bot.test.js
```diff
@@ -124,7 +139,7 @@ describe('Bot', function () {
     const text = 'testmessage';
     const newConfig = { ...config, ircNickColor: false };
     const bot = new Bot(newConfig);
-    const guild = createGuildStub();
+    const guild = this.guild;
```
ekmartin: maybe just search replace the usages of `guild` to `guild: this.guild` instead? alternatively you could do `const { guild } = this;` - but might as well just inline the usages.

test/bot.test.js
```diff
@@ -20,24 +20,50 @@ describe('Bot', function () {
...
 
     ClientStub.prototype.say = sandbox.stub();
     ClientStub.prototype.send = sandbox.stub();
     ClientStub.prototype.join = sandbox.stub();
     this.bot = new Bot(config);
     this.bot.connect();
+
+    this.addUser = function (user, member = null) {
+      const userObj = new discord.User(this.bot.discord, user);
+      const guildMember = Object.assign(member || user, { user: userObj });
```
ekmartin: This will mutate either `member` or `user`, might want to do `Object.assign({}, member || user, { user: userObj })` instead or use object spread (not that it matters too much, but mutating arguments leads to funky bugs from time to time)

Throne3d:
I've addressed the changes in this commit – thanks!

ekmartin:
Thanks - sorry about the delay!

Issue #237, opened 2017-05-12
IRC to Discord: ISO 8859-1 encoded umlauts (at least ä, ö) are corrupted


rnyberg:
Some ISO 8859-1 output from old IRC clients gets pushed into Discord like this:

`[3:44 PM] BOT IRC: <User> ny l�htis l�mmitt��n saunan`

rnyberg:
Any plans to do something to this? It does look quite nasty and it's difficult to get people to change their encodings as most IRC clients are capable of reading messages both in ISO 8859-1 or UTF-8 in the same channel, heuristically autodetecting which encoding is used.

"It's worked for years, you are the only one who complains."

Throne3d:
I'm not particularly sure where the issue would be arising in the process but it's plausible it'd be in our IRC library, which unfortunately doesn't appear to be very well maintained and as such it may be a while before it's fixed and would be an upstream issue. It might *not* be that it's a problem with `node-irc`, but if it is, it'd go with the list of other problems associated with that library. I can't currently investigate as I'm busy but I'll try to remember to do so in about a week or two, if nobody else does first?

Throne3d:
I don't suppose you could give me a server this happens on, or a client that would produce this problem? It might help track it down. It also seems like a thing to be fixed in the upstream library, [irc-upd](https://github.com/Throne3d/node-irc), which I've now taken over.

rnyberg:
Sure I can. I'll need to ask around a bit first, but I suspect mIRC prior to 7.0 will work right out the bat. I'll update with a reproductible testcase asap.

Throne3d:
It looks like the upstream library actually converts these (and displays them) just fine if you add `"encoding": "utf-8"` to the `ircOptions` object (and have `iconv` and `node-icu-charset-detector` installed). At least, it did when I tried it and mIRC with Unicode mode disabled (and the characters appeared corrupted when I removed the encoding option from the config).

Could you try that and get back?

rnyberg:
Will test.

Merri:
Fixes the issue. Should be the default setting.

rnyberg:
Fixed.

The three affected users my channel had are no longer pushing � characters into Discord.

Throne3d:
If it's the default setting it seems likely to cause some problems with how often it fails to install. I'll suggest a note in the README, instead, then users can set it up themselves?

[...]

PR #315, opened 2017-09-30
Set default encoding to UTF-8

Throne3d:
Per #237, it would be good to default to using the UTF-8 encoding when the upstream IRC library is able to support this. There should also be some sort of mention in the README about how to ensure the bot can convert encodings.

This doesn't have tests for these lines. I'm not sure how to stub `irc.canConvertEncoding()` except by including a new development dependency of `proxyquire`, which is designed for specifically the purpose of stubbing requires on other files. Suggestions, comments?

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -92,11 +92,12 @@ class Bot {
     };
 
     // default encoding to UTF-8 so messages to Discord aren't corrupted
-    if (!('encoding' in ircOptions)) {
+    if (!Object.prototype.hasOwnProperty.call(ircOptions, 'encoding')) {
       if (irc.canConvertEncoding()) {
         ircOptions.encoding = 'utf-8';
       } else {
-        logger.warn('Cannot convert message encoding; you may encounter corrupted characters with non-English text.');
+        logger.warn('Cannot convert message encoding; you may encounter corrupted characters with non-English text.\n' +
+          'For information on how to fix this, please see: https://github.com/Throne3d/node-irc#character-set-detection');
       }
     }
 
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -91,6 +91,15 @@ class Bot {
       ...this.ircOptions
     };
 
+    // default encoding to UTF-8 so messages to Discord aren't corrupted
+    if (!('encoding' in ircOptions)) {
+      if (irc.canConvertEncoding()) {
+        ircOptions.encoding = 'utf-8';
+      } else {
+        logger.warn('Cannot convert message encoding; you may encounter corrupted characters with non-English text.');
+      }
+    }
+
     this.ircClient = new irc.Client(this.server, this.nickname, ircOptions);
     this.attachListeners();
   }
--- a/README.md
+++ b/README.md
@@ -100,7 +100,8 @@ To retrieve a discord channel ID, write `\#channel` on the relevant server – i
 
 ### Encodings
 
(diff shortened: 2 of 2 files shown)


ekmartin:
I might be misunderstanding something here, but isn't utf8 already the default encoding in node-irc? https://github.com/Throne3d/node-irc/blob/master/lib/irc.js#L866-L868

If so I guess you'd only need the check and the warning message.

lib/bot.js
```diff
@@ -91,6 +91,15 @@ class Bot {
       ...this.ircOptions
     };
 
+    // default encoding to UTF-8 so messages to Discord aren't corrupted
+    if (!('encoding' in ircOptions)) {
+      if (irc.canConvertEncoding()) {
+        ircOptions.encoding = 'utf-8';
+      } else {
+        logger.warn('Cannot convert message encoding; you may encounter corrupted characters with non-English text.');
```
ekmartin: You might want to link to the section in your node-irc README here.

ekmartin:
Hm, nevermind: https://github.com/Throne3d/node-irc/blob/master/lib/irc.js#L1277-L1279

lib/bot.js
```diff
@@ -91,6 +91,15 @@ class Bot {
       ...this.ircOptions
     };
 
+    // default encoding to UTF-8 so messages to Discord aren't corrupted
+    if (!('encoding' in ircOptions)) {
```
ekmartin: `if (!ircOptions.encoding)` or if you really want to explicitly check that the property isn't set: `!ircOptions.hasOwnProperty('encoding')`

Throne3d: The linter seems to complain about `ircOptions.hasOwnProperty('encoding')`:

```
.../discord-irc/lib/bot.js
  95:21  error  Do not access Object.prototype method 'hasOwnProperty' from target object  no-prototype-builtins
```

I've made it use `Object.prototype.hasOwnProperty.call(ircOptions, 'encoding')` instead.

I think I decided to check for the property because if `encoding` is set without support for translating encodings, it will output errors for each message it attempts (and fails) to convert the encoding of. So, if the library erroneously says it can convert an encoding, and so `discord-irc` sets the encoding, (or if something else goes wrong along the way) it could be useful for people to manually disable the option by setting it to null.

lib/bot.js
Throne3d: Good idea! Done.

lib/bot.js
Throne3d: Also – what's the difference between `'encoding' in ircOptions` and `ircOptions.hasOwnProperty('encoding')`? Whether the object `ircOptions` implements it itself, versus whether some superclass of it does?

ekmartin: Yep: https://stackoverflow.com/questions/13632999/if-key-in-object-or-ifobject-hasownpropertykey

Probably doesn't really matter here though, but I think in general `hasOwnProperty` is considered a bit more hygienic.

lib/bot.js
```diff
@@ -91,6 +91,16 @@ class Bot {
       ...this.ircOptions
     };
 
+    // default encoding to UTF-8 so messages to Discord aren't corrupted
+    if (!Object.prototype.hasOwnProperty.call(ircOptions, 'encoding')) {
```
ekmartin: Hm, why did you use `Object.prototype.hasOwnProperty` instead of directly on `ircOptions`?

Throne3d: The linter complained – I said in [this comment](https://github.com/reactiflux/discord-irc/pull/315#discussion_r142007077).

ekmartin: ah lol, sorry about that :)

PR #321, opened 2017-10-04
Support displayUsername when sending message to discord

anirbanmu:
This allows displayUsername to be used in the command prelude. Before this change, if displayUsername was used in command prelude, it would not display correctly when a command was sent from IRC to Discord.


ekmartin:
Thanks!

PR #325, opened 2017-10-11
Fully implement ability to ignore IRC clients and Discord users.

### LINKED ISSUE #324: Ignoring IRC user host
How do we make this ignore an irc user host instead of nick

lalbornoz:
Fully implements the ability to ignore IRC clients by regular expressions applied against masks (sans nickname) and Discord users by nickname.
Fixes #324.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ First you need to create a Discord bot user, which you can do by following the i
     "commandCharacters": ["!", "."],
     "ircStatusNotices": true, // Enables notifications in Discord when people join/part in the relevant IRC channel
     "ignoreUsers": {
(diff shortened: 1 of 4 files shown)


Throne3d:
This looks like it would break backwards compatibility, in that using IRC nicknames would no longer function for the ignore list, wouldn't it? (Since it checks the host, and requires we match the whole part after the username.) If so, I think we can't merge it until a major version release, which will probably block this for a while.

Throne3d:
I'm not comfortable merging something that breaks backwards compatibility, and this hasn't been responded to in a long time. Accordingly, I'm gonna close it for now. If you have ideas for how to implement this in a backwards-compatible way, please reopen this!

PR #329, opened 2017-10-22
Test against Node 8, update package-lock.json

Throne3d:
Node 8 has been out for a while, and should be supported as it seems likely a lot of the userbase will be using it. This PR adds it to the list of Node versions Travis tests, to make sure there are no idiosyncracies between the versions.

This PR also updates `package-lock.json` per recent changes to the `package.json` file. Travis looks like it might be stalling while having trouble running `npm install` (a problem parsing some JSON?) so this might fix that too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,4 @@ after_success:
 node_js:
   - '6'
   - '7'
+  - '8'
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,4 @@ after_success:
   - npm run report
 node_js:
   - '6'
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -4,3 +4,4 @@ after_success:
 node_js:
   - '6'
   - '7'
```
ekmartin: Let's remove v7? Since it's not an LTS release.

Throne3d: That seems reasonable; done.

ekmartin:
Thanks 👍

Issue #246, opened 2017-05-30
changing the status of the discord bot


Rhodesia:
title says it all, is there a way to change the status message of the bot?
ex: playing xxxx

LambArchie:
Looking quickly through the code, the code would need to be inserted around [here](https://github.com/reactiflux/discord-irc/blob/master/lib/bot.js#L94). The docs on discord.js give an example [here](http://discordjs.readthedocs.io/en/latest/examples.html#status-updates). With a minor modification it could probably get details from the config file and maybe have an option to say its streaming instead.

ekmartin:
Happy to merge a PR that does this behind a `discordStatus` configuration option.

Throne3d:
All the current PRs for this (#423, #368, #359) are stalled pending consensus on an implementation and/or tests. If you have input on what specific functionality you'd like available, please get it known so we can try to add the feature!

PR #359, opened 2018-02-02
Add config option discordStatus

nindwen:
See issue #246. I needed this on my server, and this seems to work.

It might be more powerful to expose the whole [PrecenseData](https://discord.js.org/#/docs/main/stable/typedef/PresenceData) object as a setting, but I think simple string is nicer to use and covers most use cases.


Throne3d:
<https://github.com/reactiflux/discord-irc/pull/423#issuecomment-519559618>

Any idea of a consensus between this, #423 and #368?

PR #368, opened 2018-02-17
Add Presence System

nickwanninger:
This allows the bot to show their presence to the discord guild as "WATCHING irc://server/+port

This only runs when the .alterPresence option in the config is truthy


Throne3d:
<https://github.com/reactiflux/discord-irc/pull/423#issuecomment-519559618>

Any idea of a consensus between this, #423 and #359?

PR #388, opened 2018-04-20
Fix parsing of animated emoji from Discord

### LINKED ISSUE #384: Better parsing emoji
Hello,

Some emoji are printed like that on IRC : `<a:in_love:432887860270465028>`
I think it should be : `:in_love:`

Throne3d:
(Fixes #384.)

On the Discord side of things, the new animated emoji get `a` prefixed to their identifier string, so that instead of the normal emoji string:
`<:test:330064465703862273>`
you instead get:
`<a:test:330064465703862273>`

This commit fixes `parseText` to also process the animated form into `:test:` (as we currently do for non-animated emoji).

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -275,7 +275,7 @@ class Bot {
         if (role) return `@${role.name}`;
         return '@deleted-role';
       })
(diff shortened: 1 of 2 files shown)


ekmartin:
Great!

PR #419, opened 2018-07-04
Add format.webhookAvatarURL option (fixes #383)

### LINKED ISSUE #383: Use avatar for IRC user
Hello,

When user is only on IRC and not at all on discord they don't have avatar.
Maybe we should use "http://avatars.adorable.io/" for this case ?

Geo1088:
As discussed in #383, adds a `format.webhookAvatarURL` option to the config which allows users to point to a custom URL or web service to generate avatars for unrecognized IRC nicks in Discord. The option is left undefined by default, which is handled by not sending an avatar at all (which is the current behavior). I'm open to changing this default if that would be better.

I attempted to add a test for this, but it appears there is a discrepancy in all the tests between `this.bot` and individually created `bot` variables, which prevents `addUser` from working on bot instances created for individual tests. I worked around this by simply setting `this.bot = bot` in the new test, but I have a feeling this isn't a good practice (I'm new to writing tests in general). Further guidance on that would be appreciated. I would be willing to attempt another PR to correct this behavior across all tests if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,10 +76,11 @@ First you need to create a Discord bot user, which you can do by following the i
       "commandPrelude": "Command sent by {$nickname}", // Message sent before a command
       "ircText": "<{$displayUsername}> {$text}", // When sending a message to IRC
       "urlAttachment": "<{$displayUsername}> {$attachmentURL}", // When sending a Discord attachment to IRC
-      "discord": "**<{$author}>** {$withMentions}" // When sending a message to Discord
+      "discord": "**<{$author}>** {$withMentions}", // When sending a message to Discord
       // Other patterns that can be used:
       // {$discordChannel} (e.g. #general)
       // {$ircChannel} (e.g. #irc)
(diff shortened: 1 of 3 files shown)


Throne3d:
This looks good! I am also a little concerned by the discrepancy in the tests – I think I added some of the functions around this, so I guess I contributed to the problem. Probably we could pull out these three lines, somehow, for use in tests that use a non-default config?

https://github.com/reactiflux/discord-irc/blob/ef553d37e25e8437f52e3699a4aaa83b48d22708/test/bot.test.js#L39-L41

(I realize it's been a while!)

Geo1088:
Thanks for the reply! I'm no longer using this project actively, but looking over this again, it looks like the options for making the tests cleaner are either make every test init its own bot instance, or move the tests that use custom configurations to a new file and let them do their own thing where the existing `beforeEach` doesn't impact them. That seems out of scope for this PR though, and someone better acquainted with the project should probably handle that.

Throne3d:
Tested this manually a little and it works nicely! Gonna merge it. Thanks for your contribution :D

PR #421, opened 2018-07-20
Docker

gdude2002:
This pull request adds a number of things.

* A `Dockerfile`, for building Docker containers
* A `docker-compose.yml`, for easily spinning up the Docker container
* A short section in `README.md` about using Docker
* A CI step for building and deploying the Docker container to Docker Hub, including a shell script

---

**A note about CI:** If you accept this PR, please note that you'll need to add the `DOCKER_USERNAME` and `DOCKER_PASSWORD` environment variables to Travis before it'll be able to push containers to Docker Hub. These correspond with the login credentials for a Docker Hub account. I suggest you do this before squashing and merging the PR, as the build script will not deploy the container if the Dockerfile hasn't been modified.


Trojaner:
@ekmartin Can you merge and deploy this please.

docker/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN apk add --update tini
+
+RUN mkdir /bot
+COPY . /bot
+
+WORKDIR /bot
+
+ENV LIBRARY_PATH=/lib:/usr/lib
+
+RUN npm install
```
3stadt: Hey @gdude2002! :)

I'd suggest you use fewer layers here, like so:

```
FROM node:8-alpine
ENV LIBRARY_PATH=/lib:/usr/lib

RUN mkdir /bot
COPY . /bot

WORKDIR /bot

RUN apk add --update tini && \
	npm install && \
	npm run build && \
	mkdir /config

ENTRYPOINT ["/sbin/tini", "--"]
CMD ["npm", "start", "--", "--config", "/config/config.json"]
```

spacemud:
Would be great to see this merged!

gdude2002:
It's been over a year now, so I have a feeling this isn't likely to happen, sadly. Still, we can hope!

Throne3d:
Unfortunately I have very little experience with Docker in practice, so I'm not confident merging this PR by myself. I want to clarify a couple of things anyway, but these may well be introductory-level Docker questions:

1. Putting the config file in `/srv/discord-irc` seems concerning: that's a bit of a weird directory, and it's on the host, right? It's not where I'd expect this app to be making changes. It seems like it would be better to instead use a path relative to the folder the git repo has been checked out into. (Do we need an absolute path for some reason?)
2. Is it standard practice to push a `discord-irc:latest` on each change to the master branch of the git? I would have expected that to be the latest _stable_ release. (Should there also be a `discord-irc:stable`?)
3. Related to 2, should we also note to push versioned tags in our release process? Is this just an MVP, designed to get the docker image out and working asap?

3stadt:
Hey @Throne3d,

those questions are perfectly fine!
I just helped a bit with the Dockerfile itself and did not look at the rest of the PR yet, but maybe I can help a bit.

1. The config file is indeed created on the host in `/srv/discord-irc` and this is weird in my book as well. In fact, I would recommend not using a compose file at all and document how to use docker directly. Compose is useful if you need to orchestrate multiple containers; this is not done here.

2. (and 3) The latest image is usually not stable, but the latest code that was put online. This way, docker image users have the means to live on the edge as well. :D Because of that, it is not considered using the `latest` image in production. Usually for each tag as well, so in your case, there would be a docker image `reactiflux/discord-irc:2.7.0`. Lastly, I don't think a separate bash script is needed here (anymore?) since building and pushing docker images is something that can be done easily in the `.travis.yml` file directly as [documented here](https://docs.travis-ci.com/user/docker/).

At least that's my two cents on this discussion; I hope it helps! 😊

Greetings, 3stadt

gdude2002:
Hi hi, thanks for the questions.

1. Compose files are usually intended as example files when building infrastructure, but they are also an easy way to get a container up quickly. `/srv` is not an unusual place for application configs to be placed - saltstack, atlassian products and some webservers come to mind.

  I  included the compose file because it's a very convenient way to get a container up quickly. I also use them in production myself - it's quite nice to be able to grab a compose file and just modify the mounts and version.

2. It is standard practise to push the latest git tip to the `latest` tag, yes. Some projects prefer to use the latest versioned release for this and to push a dev tag and a tag that matches the commit SHA. Whatever you prefer is up to you, though.

3. If you can, then yes, you should push veesion-tagged releases. I just designed this to work with the existing CI, you could also use the CI on dockerhub to build the containers if that's more your thing.

I originally created this PR when I had a bit less Docker experience. We could clarify in the README that the compose file is intended to be edited so the config can go wherever the user wants - but it's worth noting that Docker will be working with the file as root, if discord-irc ever ends up modifying it.

Throne3d:
1. Could we use a volume mount in a command, and instead instruct the user to change the path to the config file there? Since this only uses the volume for a single config file, that seems better than making a whole separate folder for it. Also, as far as I know, discord-irc never modifies the file, so we shouldn't encounter issues around root ownership (and besides, if we did, the user running the command should be trusted as an admin anyway, per the warning in <https://docs.docker.com/install/linux/linux-postinstall/#manage-docker-as-a-non-root-user>).
2. That makes sense! I'd be alright having a `discord-irc:latest`, since that's typical practice.

To get this to work, it seems good to follow the Travis CI example more closely. (The current script looks to fail under `sh` anyway, per <https://travis-ci.org/reactiflux/discord-irc/jobs/564873707#L651>, since it can't "find" `[[` – should be run under `bash`, I assume.) If we move to the `deploy` section of `.travis.yml`, then we also shouldn't be trying to duplicate this work after each build success (right?), just after the whole build matrix is done. I'm not sure if we'll also need to modify other bits of the Travis file to make it work, but maybe (sort of like this <https://docs.travis-ci.com/user/build-stages/share-docker-image/> but inverted?).

gdude2002:
We could use a volume mount in a command, yep. I find that this is easier for beginners to grok, but yeah, there's no problem with that.

I find that mounting a file directly instead of a folder can cause some weird problems - it might work fine with discord-irc, but I've worked with other projects where it just causes cryptic errors. I guess we'd need to check.

Honestly though, it's been a long time and I have memory issues, I can't remember my entire rationale for structuring the PR this way.

I'm not super familiar with Travis. The script I provided is modified from a script I used in another project (where it did work) - it's definitely possible that Travis' environment is different, though. If you're more familiar with Travis then you could fork my fork - or if you have direct repo access, you should be able to commit directly, I have "Allow edits from maintainers" checked.

Throne3d:
I'm prodding at documentation to try to figure out the correct way to do this, but I stumbled across a thing in the script I'd missed before: it's intended to skip if we don't modify Dockerfile? That seems weird, to me – lots of updates to the codebase (which will then affect the docker image) won't modify the Dockerfile, and in fact the only things that _would_ affect the Dockerfile are things like changing the node version or other dependencies.

Is it fine to take that out?

gdude2002:
Oh, right you are. I must have mistakenly left that in - I'll go edit it now.

Throne3d:
A few comments:

1. Looks like running `docker-compose up` is trying to fetch `reactiflux/discord-irc` from Docker Hub (which doesn't currently exist). Since this is only ever going to be used with a cloned git repo, fetching from Docker Hub seems wrong, so I'm gonna go ahead and remove the docker-compose file for now, though if people happen to use the image with other services a lot (I don't think they do), we could later provide a sample file based on this one.
2. It looks like, based on where the Dockerfile is located, simply `cd`-ing to that directory and building it fails (because of the `COPY` command using `.`, locally). I'm going to move it to the root to solve this, though you might also be able to use `-f` on a `docker build` command.

I'm attempting to fix the Travis script as well, but I'm not sure it'll work. I'm also not super sure about docker image naming conventions, but I think the stuff I've put in the README should work.

This should cause Travis not to run the deploy script, but on merge, it should end up trying to. I'm planning to try it in a fork. It looks like I can't push to your branch (though I am able to merge PRs in this repository).

Throne3d:
Here's the fork I put it in, fwiw: <https://github.com/Throne3d/discord-irc>. The Travis script seems to work alright! It took a bit of fiddling to get it to successfully skip all of the deploy phase on non-master branches (using the condition in the `deploy` section made it still try running, it just ended up skipping the actual deploy phase), but I succeeded. The master branch also successfully runs it and builds the image, then tries to push (and fails because I haven't given it credentials).

Looks like a success!

Throne3d:
… Shit, I pushed to the wrong repo. Hang on a sec.

gdude2002:
Haha, I figured you merged it to this repo intentionally. Whoops.

The script _should_ be okay, I don't recall there being a `cd` at all. It's based on one I used to build this repo myself, in a private GitLab repo.

image

I've been using compose to run this for a year now on my own containers!

Throne3d:
(The `cd` was in the README before I amended it in <https://github.com/reactiflux/discord-irc/commit/977b337a53efa8a67e3cac790dc56d6092ff1a1d#diff-04c6e90faac2675aa89e2176d2eec7d8R42>, and it made the `docker-compose` script break.)

Should now be good in #498! I'm happy with what's now there.

PR #423, opened 2018-07-22
Add config option for Discord setActivity

Golumpa:
Another alternative to resolve issue #246 

I chose to use the setActivity class as in my opinion, makes more sense for something like this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,8 @@ First you need to create a Discord bot user, which you can do by following the i
       "irc": ["irc_nick1", "irc_nick2"], // Ignore specified IRC nicks and do not send their messages to Discord.
       "discord": ["discord_nick1", "discord_nick2"] // Ignore specified Discord nicks and do not send their messages to IRC.
     },
+    // Sets the discord activity status, can be set to one of "PLAYING, STREAMING, LISTENING or WATCHING"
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -90,6 +90,8 @@ First you need to create a Discord bot user, which you can do by following the i
       "irc": ["irc_nick1", "irc_nick2"], // Ignore specified IRC nicks and do not send their messages to Discord.
       "discord": ["discord_nick1", "discord_nick2"] // Ignore specified Discord nicks and do not send their messages to IRC.
     },
+    // Sets the discord activity status, can be set to one of "PLAYING, STREAMING, LISTENING or WATCHING"
+  	"discordStatus": ["PLAYING", "with you!"],
```
ekmartin: Maybe it would be more self explantory if this was an object? E.g. something like:

```js
"discordStatus": {
  "name": ...,
  "type": ...,
  "url": ...
}
```

Throne3d:
I can't tell what's going on between #423, #368 and #359. Are there any plans to change the PR further, and would you be able to add (basic) tests for the functionality? (Is there consensus on a specific implementation?)

I don't care much about the rich presence system, so something that people agree works would be good.

Issue #434, opened 2018-09-14
Specify irc port in config?


nand2xor:
In the standard example config there's only the server address/name. I see no parameters to add any custom port. The server I try to connect to only uses port 6697 with ssl.

So when I try to run my discord bridge, it only connect to the discord. And then nothing happens, no output or information whether or not irc is connected.

Is there any more detailed documentation for the config.json?

Mikaela:
https://github.com/reactiflux/discord-irc/issues/425#issuecomment-408351588 (and https://github.com/reactiflux/discord-irc/issues/308 & https://github.com/reactiflux/discord-irc/issues/76)

ekmartin:
Thanks @Mikaela — `ircOptions.port` is the way to go.

Issue #426, opened 2018-08-02
Auth and +r channels


AcidWeb:
Bot don't wait for end of authorization process and that makes it impossible to use with +r channels.

Mikaela:
Have you thought about enabling SASL in your bots so they get identified to NickServ before getting connected to the network? Sasl, username and password from https://node-irc-upd.readthedocs.io/en/latest/API.html#irc.Client if I recall correctly.

Another option would be `mode +b $~a` to ban unidentified users and `mode +e` the bot.

AcidWeb:
SASL is correct answer to this problem.

Issue #308, opened 2017-09-25
SSL network


scramble0:
Trying to connect to `hub.wirelesspt.net SSL 6697`
So it requires on "ircOptions":  `      "port": 6697,
      "secure": true,
      "selfSigned": true,
      "certExpired": true `

But I get the following error on console

> info: Connected to IRC
/usr/lib/node_modules/discord-irc/node_modules/irc/lib/irc.js:849
                        throw err;
                        ^
TypeError: Cannot read property 'add' of undefined
    at Client.ircClient.on (/usr/lib/node_modules/discord-irc/dist/bot.js:157:61)
    at emitThree (events.js:135:13)
    at Client.emit (events.js:216:7)
    at Client.<anonymous> (/usr/lib/node_modules/discord-irc/node_modules/irc/lib/irc.js:495:22)
    at emitOne (events.js:115:13)
    at Client.emit (events.js:210:7)
    at iterator (/usr/lib/node_modules/discord-irc/node_modules/irc/lib/irc.js:846:26)
    at Array.forEach (<anonymous>)
    at TLSSocket.handleData (/usr/lib/node_modules/discord-irc/node_modules/irc/lib/irc.js:841:15)
    at emitOne (events.js:115:13)

on IRC

`botname (bot@address) has quit IRC (Read error)`

help? thank you!

Throne3d:
Uh, just to check, are you on an old version? It looks like that's using the `irc` library (instead of `irc-upd`, the newer one).

scramble0:
Toying arround  and deployed a new instance for this network, didn't check for updates! that fixed it.
Thanks for the awesome super reply @Throne3d

Throne3d:
No problem! If it happens again, it could be an actual problem – the code around there has been a little intermittent – but hopefully it won't.

Issue #76, opened 2016-06-04
Any way to connect using SSL?


Gachl:
Hi

I'm trying to connect discord-irc to a bouncer which runs SSL. Is there any way to do this? I have tried using "server": "localhost:+20001" which didn't seem to work.

```
> discord-irc --config discord.conf
debug: Connecting to IRC and Discord
debug: Connected to Discord
```
and then nothing happens

Thanks,
Gachl

edit: sorry I'm stupid, I didn't look at the available options site.

BeauBouchard:
I also am wondering this. Is there a way to connect with SSL?

Gachl:
Check the secure option here:

http://node-irc.readthedocs.io/en/latest/API.html#irc.Client

Peppernrino:
> "secure": "true",
"port": "6697"

Throne3d:
Better documentation link: https://node-irc-upd.readthedocs.io/en/latest/API.html#irc.Client

But yeah, you just set secure to true and make sure you set the correct port if applicable.

Peppernrino:
lol. i was trying to figure it out at the time as well.

PR #436, opened 2018-09-15
README.md: add SSL & SASL examples

Mikaela:
I think this is a frequently asked question on the issue tracker based on #434, #308, #76 and #426.


Mikaela:
Thanks and sorry, I was a bit sleep deprived at that time and apparently I am still not awake enough to notice that you merged it already and would have fixed it just now.

ekmartin:
No worries! Thank you for the help.

Issue #430, opened 2018-08-25
Feature Request: Block predefined spam messages


Kage:
### Problem

Our IRC network uses the UnrealIRCd `spamfilter` feature to block various defined spam messages. Currently users in Discord can send these blocked messages which either disrupts the bot (gets gzlined) or spams them (if something like an oper override is applied to the bot).

### Request

**As** an IRC administrator
**I want to** prevent Discord from sending certain predefined messages
**So that** my bot does not conflict with my `spamfilter` settings

Optimally, I would like to set an array of regular expressions in my `config.json` that, if matched, do some defined action (much like the IRC `spamfilter`). For now it's sufficient for me if that action is simply not sending the message to IRC. But other options could include Discord punishments (muting, banning, etc.). Bi-directionality (IRC users sending messages unwelcome in Discord, kicking or banning in IRC, etc.) could be useful, too.

Throne3d:
There's a PR (#455) to implement this, but it doesn't go as general as regular expressions. As with some of the other block/ignore PRs & issues (see also #324), I'm not sure what would be a good compromise between complexity and flexibility here, especially keeping in mind not breaking backwards compatibility. It would be great to get some more input before merging to figure out a good, general solution.

PR #455, opened 2018-12-13
add ignoredWords feature (doesn't forward messages containing an igno…

intrand:
…redWord)

example configuration option (top level JSON key):
"bannedWords": ["pleasebanme", "testsomewords", "thingsYouWouldntPutInACommitMessage"]

tested working, built in docker.

feel free to change code style, naming, whatever you like.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -31,7 +31,8 @@ class Bot {
     this.nickname = options.nickname;
     this.ircOptions = options.ircOptions;
     this.discordToken = options.discordToken;
-    this.ignoredWords = options.ignoredWords || []; // default don't ignore any words
+    this.filterWords = options.filterWords || []; // default don't filter any words
+    this.filterPrefixes = options.filterPrefixes || []; // default don't filter any prefixes
     this.commandCharacters = options.commandCharacters || [];
     this.ircNickColor = options.ircNickColor !== false; // default to true
     this.channels = _.values(options.channelMapping);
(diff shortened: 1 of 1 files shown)


intrand:
We quickly came to the conclusion that ignoring certain words is great, but insufficient. I will add to the merge request one more change: We added ignored prefixes, too. Since now we have more than one type of thing we filter out, we changed the wording to be uniform and a little more plain about what it does.

In the case of prefixes: Instead of making commands go straight through (like commandPrefix), we want to avoid sending (filter) any message that starts with a particular string.

"filterWords": ["listof", "verybad", "wordsthatarebad"]
"filterPrefix": ["!", "?", "."]

Code of an extremely similar style to follow in the next commit/push. I will add it to this merge request if I can.

intrand:
Have you had a chance to take a look at this?

Johj:
I think adding some granularity to a filters feature would be nice. Something like a hard vs soft mode. Hard would not send at all while soft would regex replace the filter word with a preset like "<censored>".

intrand:
if you want to extend it further, go ahead. We need the capability to completely drop messages with banned words and commands, which have different criteria (contains vs starts with). We don't like the idea of putting "<censored>" or similar in there, but we don't care if there's an option for it -- we just won't use it, hence why I don't want to spend time making it work well.

Throne3d:
Before merging this, I'd want some tests added to the PR – however, it seems like people want a bunch of different ways to ignore people from being relayed by the bot, and I'm not sure what the full set is. I'd like to get some more information on that (and maybe some more general ideas about what features to merge) before getting this in, because we'll want to preserve backwards compatibility once we add this. (See <https://github.com/reactiflux/discord-irc/issues/430#issuecomment-519564514>.)

MikkelDK:
Any ETA about this? @Throne3d @intrand ???

I did test it but seems it doesn't work with "ignoreWords" as an example with "listof"

I can still see it posted on discord.

Throne3d:
Hey! I'm afraid this PR is unlikely to be merged in the foreseeable future. I'm only now maintaining discord-irc for bug fixes and vulnerabilities, as I'm no longer an active user of the project.

I appreciate the contribution, and hope it's useful to anyone who comes across it. If you have the time to maintain a fork, that might be a way to help keep this project developing.

PR #457, opened 2019-01-01
Add webhookDisableEveryone option

tadeokondrak:
On Discord, webhooks automatically have mass ping permissions that can't be removed. This patch adds a config option to tell discord.js to insert an invisible character in mass pings to prevent this from happening.


AlmightyFloppyFish:
Any chance of this getting merged?

Throne3d:
I've implemented this by deferring to the discord permission in #497 – that way, we don't need a global configuration option; users can configure it using the built-in Discord interface and control it with the usual permissions setup in the normal way. Hopefully we can merge that soon!

PR #470, opened 2019-02-11
Match IRC-style mentions at beginning of message

rdb:
On my IRC channel I've noticed people expect to be able to address Discord users like they usually do on IRC, in one of these two forms:
```
<ircuser1> discorduser: yes, I agree
<ircuser2> discorduser, I also agree with that
```

This PR makes it possible by extending the mention regex to transform a `discorduser:` or `discorduser,` at the beginning of a message into `@discorduser`.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -476,7 +476,9 @@ class Bot {
       if (user) return user;
 
       return match;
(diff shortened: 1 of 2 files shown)


rdb:
Added unit tests; looks like the checks passed.

ekmartin:
Excellent, thanks!

PR #476, opened 2019-03-13
Add channel mentions, similar to user mentions

### LINKED ISSUE #460: Parse channels similar to how names are parsed
`@JohnDoe#1337` is turned into a user mention, channels (`#example`) should be turned into channel links.

Throne3d:
Fixes #460.

- Convert channel mentions (starting with #, ending with a
  space, #, @, ', !, ?, , or .) to actual codified mentions
- Modify test stubbing to allow for multiple guilds in tests
  (and fix an unused part of the addRole and addEmoji stubs)
- Test that the bot converts channel mentions appropriately:
  * does not convert unmatched channels
  * converts even if a trailing comma is found
  * converts even if special characters (e.g. `é`) are used
  * converts irregular case mentions
  * does not convert mentions to channels from other guilds

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -471,7 +471,7 @@ class Bot {
       // skips usernames including spaces for ease (they cannot include hashes)
       // checks case insensitively as Discord does
       const user = guild.members.find(x =>
-        Bot.caseComp(x.user.username, username.toUpperCase())
+        Bot.caseComp(x.user.username, username)
         && x.user.discriminator === discriminator);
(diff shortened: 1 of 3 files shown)


Throne3d:
Hmm, hang on, I'm not sure this actually works in practice.

Throne3d:
Oh, nope, works fine, I just forgot to transpile. Haven't worked on this project in a while.

ekmartin:
Great 👍

PR #496, opened 2019-07-26
fix: prevent using @everyone from an IRC channel

### LINKED ISSUE #494: IRC users can ping @everyone
Users on IRC can type `@everyone` to ping everyone, even when this is disabled in all the roles that a bot is a member of.

## User

image

## Roles

* User is a member of Bot and Everyone, both have mention everyone disabled.
* The channel has mention everyone disabled on both Bot and Everyone.
* None of these roles have admin privileges set

### Bot role

image

me4502:
This prevents IRC users spamming @everyone in massive Discord chats :) 

Fixed #494

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -560,7 +560,8 @@ class Bot {
       webhook.client.sendMessage(withMentions, {
         username,
         text,
-        avatarURL
+        avatarURL,
+        disableEveryone: true
       }).catch(logger.error);
       return;
     }
--- a/test/bot.test.js
+++ b/test/bot.test.js
@@ -986,6 +986,7 @@ describe('Bot', function () {
       username: 'n_',
       text,
       avatarURL: null,
+      disableEveryone: true,
     });
   });
 
@@ -999,6 +1000,7 @@ describe('Bot', function () {
       username: '12345678901234567890123456789012',
       text,
       avatarURL: null,
+      disableEveryone: true,
     });
   });
 


Throne3d:
It looks like this and #457 are designed to solve the same issue – the bot lets users from IRC use `@everyone`, without that necessarily being intended.

However, since sometimes it might be desirable to let users on IRC side use `@everyone` (say, on small servers where the security doesn't matter, or where the bridge is an announcements channel), it would be nice to let users turn this on sometimes. Additionally, since we _already_ have a permission through Discord for whether the bot should be able to use `@everyone` or not, could we just disable it based on that?

We can check if the bot's got permission to use `@everyone` with:
```js
const permissions = discordChannel.permissionsFor(this.discord.user);
let canPingEveryone = false;
if (permissions) {
  canPingEveryone = permissions.has(discord.Permissions.FLAGS.MENTION_EVERYONE);
}
```

Then later we can just use this in the `sendMessage` hash:

```js
webhook.client.sendMessage(withMentions, {
  username,
  text,
  avatarURL,
  disableEveryone: !canPingEveryone,
}).catch(logger.error);
```

That way, we don't need a global configuration setting (like #457 did), nor are we blocking out flows where a bot operator might want this functionality.

At this point I suppose I could just write the PR myself, but it'd be nice to get some feedback on this idea before I just go about doing that (or you could implement it and maybe add tests for it?). :P

me4502:
Yeah if we could make it just obey the permissions provided by Discord that would make the most sense. Especially from a usability perspective, where people will just assume removing @everyone permissions prevents it from happening.

Throne3d:
I've set this up in #497!

PR #497, opened 2019-08-01
Defer to discord for disabling at-everyone in webhooks

### LINKED ISSUE #494: IRC users can ping @everyone

Throne3d:
Currently, webhooks permit IRC users to send `@everyone` and `@here` pings through, as they're not stripped (and webhooks always have this permission).

Discord already has permissions for whether the bot can ping everyone, and the webhook send method has a `disableEveryone` option, so let's defer to the inbuilt Discord permission for this.

When permissions can't be found (shouldn't happen), defaults to disabling the pings.

Fixes #494. Supersedes #496 and #457.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -555,12 +555,18 @@ class Bot {
     const webhook = this.findWebhook(channel);
     if (webhook) {
       logger.debug('Sending message to Discord via webhook', withMentions, channel, '->', `#${discordChannel.name}`);
+      const permissions = discordChannel.permissionsFor(this.discord.user);
+      let canPingEveryone = false;
+      if (permissions) {
+        canPingEveryone = permissions.has(discord.Permissions.FLAGS.MENTION_EVERYONE);
+      }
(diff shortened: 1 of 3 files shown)


Tsunder:
This appears to, when a user with a name partially matching here / everyone, for example "eve" exists in the server, will ping "eve" instead of everyone.

Throne3d:
That's standard for Discord pings – if you type `@Throne3dtest`, and there's a user `@Throne3d` on your server, it'll ping them with "test" appended (no space). I think it's fine for the behavior to stay like this, because of that.

Throne3d:
I'll merge this for now, since it seems quite important for large bridges – we can amend it later if needs be.

PR #501, opened 2019-08-05
Run all test/*.test.js files

Throne3d:
Somehow, upgrading either mocha or nyc (?) appears to make the current test command skip test/bot-events.test.js. I do not know why, but quoting it (so mocha globs it instead) seems to fix this.

(This happened between 2.7.0 and 2.7.1.)


Throne3d:
Coverage is nicely back to the high figure it was at before! I consider this a success. Merging.

PR #502, opened 2019-08-08
Add ircPreventMention

### LINKED ISSUE #197: Avoid self-message highlighting on IRC

Throne3d:
Supersedes #243. Original credit to @qaisjp, test added, README amended to state it's off by default. Single-character nicks get a random zero-width character appended, but it's not like that affects anything so meh.

Off by default, because some IRC clients were having trouble with it before. It's up to bot owners to make sure this doesn't break any functionality for their users (like maybe for languages where zero-width characters will mess up the language's script? on IRC servers supporting those characters in nicknames).

Fixes #197.


qaisjp:
Cheers

PR #511, opened 2019-09-23
Fix avatars not displaying via webhooks

Miosame:
Avatars that were 2048x2048 would break webhooks displaying avatars, defaulting to 128x128 fixes that.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -578,7 +578,7 @@ class Bot {
       if (permissions) {
         canPingEveryone = permissions.has(discord.Permissions.FLAGS.MENTION_EVERYONE);
       }
-      const avatarURL = this.getDiscordAvatar(author, channel).replace(/\?size=\d{1,}$/,"?size=128");
+      const avatarURL = this.getDiscordAvatar(author, channel).replace(/\?size=\d{1,}$/, '?size=128');
       const username = _.padEnd(author.substring(0, USERNAME_MAX_LENGTH), USERNAME_MIN_LENGTH, '_');
       webhook.client.sendMessage(withMentions, {
         username,
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -433,7 +433,7 @@ class Bot {
 
     // No matching user or more than one => default avatar
     if (users && users.size === 1) {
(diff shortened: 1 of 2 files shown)


Miosame:
Fixed all of the relevant CI issues, the remaining ones are to be fixed by you, because they expect the old 2048x2048 size:

```
AssertionError: expected '/avatars/123/avatarURL.png?size=128' to equal '/avatars/123/avatarURL.png?size=2048'
```

Throne3d:
It's very strange to claim the remaining CI issues are to be fixed by us – it's low effort to fix the tests to expect the right results, and the code change you're proposing does break old expected behavior, so the remaining issues are, relevantly, yours to fix.

Can you provide reproduction steps for the issue this fixes?

Miosame:
@Throne3d I'm not sure why the note about the tests set you off? at the time I wasn't simply aware that the tests are hosted on this repository too and noted that the tests would need to be modified for the new case, as I assumed those are hosted on travis internal configurations in some way.

I will re-test if this is still necessary as a fix and document further in the coming hours.

Miosame:
@Throne3d tested it just now and the issue still persists, here are the steps to reproduce it:

1. Post in IRC with a normal avatar that does not exceed the size => the bot displays the avatar
2. Change the avatar to a very large image, e.g.: https://w.wallhaven.cc/full/od/wallhaven-odjx85.jpg
3. Post again in IRC => bot attempts to fetch, showing the default discord avatar and then goes to a transparent image like this:

image

I will push the test fixes shortly.

Miosame:
@Throne3d submitted what looked like failing tests purely based on avatar size and now the tests pass. The above coverage comment was before I merged upstream.

Throne3d:
> @Throne3d I'm not sure why the note about the tests set you off?

I didn't mean to come across as being set off – your comment about the tests just came off as very strange =P

Thank you for this code change! It's really strange that I couldn't find anything about this in Discord's documentation, but I'm happy to merge this as-is. I'm slightly concerned that 128px might be too small, but if anyone notices a visual change, I'm hoping to get an issue and we can look into bumping it up.

PR #523, opened 2019-12-14
Upgrade to Babel 7

Throne3d:
Babel 6 was bringing a bad dependency through (in development), and Babel 7 was released a long time ago. Let's do a quick upgrade (like this) and hopefully ignore it for a while longer.


Throne3d:
No lines seem to be actually differently covered, it's just reporting different percentages in the summary. I assume this is due to the difference in transpiling, so I'll just go ahead and merge it anyway.

PR #537, opened 2020-03-20
Add setting to disable irc- or discord-style mentions

### LINKED ISSUE #528: Make IRC->Discord notifying without '@' prefix behavior configurable
This was the behavior introduced in #470, but doesn't mesh well in communities where people are using both IRC and Discord; users talking to each other in IRC will receive a bunch of Discord notifications when one highlights the other. There should be a setting to have IRC->Discord highlights always require the `@` prefix. 
```
<Discord> <Link> this new bot allows irc to discord @highlights
<Austin> i mentioned that
<Austin> and i dont like that it doesnt require @
<Discord> <Link> Try it on me
<Austin> Link, no
[appears in Discord as "@Link no"]
<Discord> <Link> Was that without @?
<Austin> yes
<Discord> <Link> Hmm
```

Gargaj:
This fixes #528 which is really annoying.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -42,6 +42,8 @@ class Bot {
     this.ircStatusNotices = options.ircStatusNotices;
     this.announceSelfJoin = options.announceSelfJoin;
     this.webhookOptions = options.webhooks;
+    this.allowDiscordStyleNotifyFromIRC = options.allowDiscordStyleNotifyFromIRC !== false;
+    this.allowIRCStyleNotifyFromIRC = options.allowIRCStyleNotifyFromIRC !== false;
 
     // Nicks to ignore
     this.ignoreUsers = options.ignoreUsers || {};
@@ -505,7 +507,12 @@ class Bot {
 
       return match;
     }).replace(/^([^@\s:,]+)[:,]|@([^\s]+)/g, (match, startRef, atRef) => {
-      const reference = startRef || atRef;
+      // override the detection via settings
+      const filteredStartRef = this.allowIRCStyleNotifyFromIRC ? startRef : null;
+      const filteredAtRef = this.allowDiscordStyleNotifyFromIRC ? atRef : null;
+
+      const reference = filteredStartRef || filteredAtRef;
+      if (!reference) return match;
 
       // this preliminary stuff is ultimately unnecessary
       // but might save time over later more complicated calculations
--- a/test/bot.test.js
+++ b/test/bot.test.js
@@ -558,6 +558,34 @@ describe('Bot', function () {
     this.sendStub.should.have.been.calledWith(expected);
   });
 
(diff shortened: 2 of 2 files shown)


Gargaj:
@CVSoft in case you still need it

Gargaj:
@Throne3d I know you no longer add code, but this is merge-ready, if you want :)

DarkSpyro003:
This seems simple enough for me - if there is enough interest it should probably be merged, although I'm not really a fan of the names of the new config options.

You should, however, update the README as well, advanced options under Example configuration.

Gargaj:
I admit I struggled with the option names; suggestions welcome? :/

DarkSpyro003:
Naming wise, try to KISS: it could be as simple as _allowMentionsWithAtPrefix_ and _allowMentionsWithoutPrefix_.

However, those names have made me start thinking and propose an alternate solution. I would love your input.

_(This does not mean I do not approve of your current PR, it solves a problem aptly)._

### This is what this current PR does:
The problem you want to solve is:
* Currently, the bot will try to resolve mentions even for normal words because mentions require no prefix

Your solutioning is:
* Introduce a configuration setting to disable mentions without a prefix

As a nice-to-have that matches with it:
* Introduce a configuration setting to disable mentions with a prefix.

### Possible alternative solution
Now, I'm just going to 'go wild' here and toss out a different possible approach - I would like your input on it as well:
Problem:
* The bot will resolve who to mention on words without a prefix, and words with prefix @, and this cannot be changed

Proposed solutioning
* Allow the configuration of prefixes that are used to resolve mentions.

I would imagine the default configuration could look something like this:
```json
{
    "mentionPrefixes": [
        "",
        "@"
    ]
}
```

### Conclusion
Just some food for thought. This would offer a lot more flexibility in the configuration instead of creating a hard-coded case for every commonly-asked for (combination of..) prefixes (or lack thereof) to mention people with.

Gargaj:
The problem is that "prefix" is misleading, the format it's using isn't just `@nick`, but also `nick:` and `nick,`.

Gargaj:
Perhaps instead of `mentionPrefixes` it should be `mentionRegexps`?

jesigloot:
To throw my apprecation into this; Needed exactly this feature, so compiled it into a docker image myself to run it. 😅

Throne3d:
Hey! Sorry to not have responded in a while – I haven't been tracking this repository very well.

I really appreciate the contribution, especially the added tests. I see that there's still some open discussion here, so I'm not going to merge it immediately. (I know it's not a huge difference either way, but still.) If this comes to a consensus I can try to get to merging it then!

Gargaj:
For what it's worth we've been running with this for one of our bots and it's working brilliantly.

PR #539, opened 2020-05-18
Add /names command to Discord

Kage:
This change adds a `/names` command to Discord so users there can see who is in the associated IRC channel.  The command and its output are only seen in Discord and not carried over to IRC.

Example output:

Screenshot

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -159,8 +159,17 @@ class Bot {
     });
 
     this.discord.on('message', (message) => {
-      // Ignore bot messages and people leaving/joining
-      this.sendToIRC(message);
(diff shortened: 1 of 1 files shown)


NightYoshi370:
Could you add the same except IRC users can check online discord users?

Throne3d:
Hi there! I'm not currently maintaining this project for new features, only bug fixes, vulnerabilities, and what's necessary to keep it functioning.

Thanks for your contribution, but I'm not likely to merge it – sorry!

RhinosF1:
Could we not have a way to send any command to IRC and even messages to bots so people could interact with IRC Bots in channels

PR #541, opened 2020-06-04
Fix parsing when the same user is mentioned more than once

Qyriad:
Before, the parser would only replace the first instance of a mention for any given user in a message.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -275,9 +275,8 @@ class Bot {
   parseText(message) {
     const text = message.mentions.users.reduce((content, mention) => {
       const displayName = Bot.getDiscordNicknameOnServer(mention, message.guild);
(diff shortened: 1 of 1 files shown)


Throne3d:
Hey there! Thanks for your contribution – I'd never thought to test this. Would you be able to rebase this branch onto master? (Our previous CI process was broken, and I've updated it to use GitHub Actions.) It'd also be great to have a test to make sure we don't have a regression with this, but I can add that if you'd like!

Qyriad:
I rebased it onto master like you asked. I took a quick look, and unfortunately I don't think I'm familiar enough with the codebase or the libraries you're using for tests to be able to write a unit test for this, though. Sorry!

Throne3d:
Awesome, thank you for the contribution! I'll see if I can get adding tests for this this weekend.

Qyriad:
Thank you!

PR #543, opened 2020-07-02
Added option to replace command prefix with alternative prefix

ivanovsaleksejs:
Sometimes you need to change a prefix of a command before sending it to IRC. For example, the IRC channel has bot that has s/ feature implemented. You can't send 's/a/b' from Discord (it's hardcoded to edit feature), instead you can send something like '!s/'. Modify your config:

`"commandCharacters": ["!", ["!s", "s"]]`

and it will replace prefix '!s' with 's' before sending.

### DIFF
--- a/lib/bot.js
+++ b/lib/bot.js
@@ -290,8 +290,22 @@ class Bot {
   }
 
   isCommandMessage(message) {
-    return this.commandCharacters.some(prefix => message.startsWith(prefix));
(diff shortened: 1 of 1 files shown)


Throne3d:
Hi there! I'm not currently maintaining this project for new features, only bug fixes, vulnerabilities, and what's necessary to keep it functioning.

Thanks for your contribution, but I'm not likely to merge it – sorry!

PR #581, opened 2021-02-05
discord-irc: show when messages are deleted

g-w1:
This allows for less confusion when someone deletes a message and you try to reply to it now knowing that it was deleted.


g-w1:
CLOSING BECAUSE STALE

PR #582, opened 2021-03-01
package: update

RhinosF1:
Tested locally using npm and seems to work with all but a missing ENV variable. Not sure what to do about that.

Updated commander & discord.js in package.json and various in package-lock including the manifest version.


RhinosF1:
Closing as I saw a few not make sense changes in package-lock too

PR #598, opened 2021-07-10
Log netError events from irc-upd

jdstroy:
Socket errors are logged using the netError event on irc-upd.  Log them, too.


aoeixsz4:
lmao, I just did the same thing and was about to make a pull-request
upvote!

PR #600, opened 2021-10-01
Add unemojify option to textify emojis, fixes #192

### LINKED ISSUE #192: Discord emotes aren't turned into text for IRC and certain language-specific characters are sometimes broken
It appears that emotes like :D don't turn from their Discord emoji to text. I don't know if it's the correct way but every time I post an emoji to Discord, it forwards the emoji to IRC and not its text. Is this intended behavior or not?

The other issue is with special language-specific characters like ä, ö and å. This is from IRC to Discord. It works sometimes, as seen here: 
image

However, sometimes it doesn't: 
image

This seems to only happen with certain users in IRC. Whenever they post a message with those characters, it's always broken. For some users it never happens. It could be because others use ISO 8859-1 and others use UTF-8 but I don't know for sure. Is there a way to fix this?

seisatsu:
Added the node-emoji library as discussed in #192, and an "unemojifyMessages" config option to enable it. Tested and appears to be working.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ First you need to create a Discord bot user, which you can do by following the i
       "sasl": true, // false by default
       "username": "test", // nodeirc by default
       "password": "p455w0rd" // empty by default
+      "unemojify": true // Textify emojis from messages send to Discord, false by default
     },
     "format": { // Optional custom formatting options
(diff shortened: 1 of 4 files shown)


seisatsu:
I probably should not have bumped the version number in package.json, sorry.

PR #632, opened 2023-08-24
Update to Discord.JS v14 (and move to TypeScript)

aronson:
I updated the application.

Fixed some bugs/annoyances too:
- No longer crashes when typing in a VC text channel in Discord
- Changed default Discord->IRC username mapping to use displayName instead of username
- Partial role matching was annoying in my testing so I guarded it with a new config, feel free to undo
- Updated Winston to fix logging on some newer node versions

Problems:
- I had to disable @ everyone pings because I'm not sure how to enable them
- If you try to bind to a non-text Discord channel by ID or name it'll crash the app. I suggest not doing this


dustwolf:
Much appreciated. I'm merging this with my own fork and testing...

dustwolf:
The merged code seems to be working. 

One change that I can see is that the "SERVER MEMBERS INTENT Required for your bot to receive events listed under [GUILD_MEMBERS](https://discord.com/developers/docs/topics/gateway#list-of-intents)." is now required whereas the old version didn't need it. Other than that, everything seems to have worked like a charm.

aronson:
Continuing to work on this project, I made a number of fixes/changes and also converted the application to typescript (which should not affect end-users).

Changes that affect end-users:
- I re-enabled the encoding I disabled in error
- I added a new config property, `allowRoleMentions`, true by default (the original behavior)
- I merged in the `/names` command PR from [Kage](https://github.com/reactiflux/discord-irc/pull/539)

Developer-relevant changes:
- I switched to yarn, but you don't have to
- I switched to tsx as the runner and removed the compilation step
- I changed the linting rules
- I installed prettier and ran it on everything
- I re-worked the overall package.json, removing now unused dependencies
- Sources are now in `src/` instead of `lib/`

I made opinionated development changes and foolishly opened this PR against my main branch instead of a contribute branch, so these changes are also in this PR. I now understand this project is not actively maintained so I don't think it will be an issue.

I attempted to remove the `GuildMembers` intent as dustwolf mentioned above, but it seems it's necessary for the bot. Without it messages will not make it through the bridge. Not sure why that changed.

